PR #119, opened 2026-03-27
fix: update `ParseEbuildVariables` regex for gentoo version suffixes

arran4:
Updated `ParseEbuildVariables` regex in `ebuild.go` to correctly parse gentoo version suffixes that have optional digits (e.g., `_beta`, `_alpha`) and added missing support for the `_pre` suffix.

---
*PR created automatically by Jules for task [1473872440762466227](https://jules.google.com/task/1473872440762466227) started by @arran4*

### DIFF
--- a/ebuild.go
+++ b/ebuild.go
@@ -167,8 +167,10 @@ func ParseEbuild(fsys fs.FS, path string, mode ParsingMode) (*Ebuild, error) {
 
 	// Always parse metadata from filename
 	vars := ParseEbuildVariables(path)
-	for k, v := range vars {
-		e.Vars[k] = v
+	if vars != nil {
+		e.Vars["P"] = vars.P
+		e.Vars["PN"] = vars.PN
+		e.Vars["PV"] = vars.PV
 	}
 
 	if mode == ParseMetadataOnly {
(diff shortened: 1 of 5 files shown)


ebuild.go
```diff
@@ -240,7 +240,7 @@ func ParseIUSE(iuseStr string) []string {
 // ParseEbuildVariables extracts PN, PV, P from the ebuild filename.
 func ParseEbuildVariables(filename string) map[string]string {
```
arran4: @jules I think we need to create a special parser for this that uses the version parser to keep this in sync and up to date. It should return a tuple or a structure too.

ebuild.go
```diff
@@ -237,25 +239,41 @@ func ParseIUSE(iuseStr string) []string {
...
+// EbuildVariables represents standard variables parsed from an ebuild filename.
+type EbuildVariables struct {
+	P  string
+	PN string
+	PV string
+}
+
 // ParseEbuildVariables extracts PN, PV, P from the ebuild filename.
-func ParseEbuildVariables(filename string) map[string]string {
+func ParseEbuildVariables(filename string) *EbuildVariables {
```
arran4: @jules sorry a map was correct here.

PR #126, opened 2026-03-27
Add support for Gentoo News-Item-Format 2.0 and metadata headers

arran4:
This PR implements GLEP 42 News-Item-Format 2.0 support. It introduces the extraction of additional headers such as `News-Item-Format`, `Translator`, `Display-If-Installed`, `Display-If-Keyword`, and `Display-If-Profile`. Furthermore, it hides email addresses from Author and Translator names when parsing.

The primary update is the capability to convert Format 2.0 news bodies—which include markdown-like syntax for lists and code blocks—into proper HTML structure (`<ul>`, `<li>`, `<pre>`, `<code>`). To support this accurately, the CSS `white-space: pre-wrap` was removed from the templates, and older news items (fallback) natively replace their text newlines with `<br>` tags to preserve visual formatting without breaking layout. Finally, the newly added headers are formatted in a neat definition list block within `news_article.html`.

---
*PR created automatically by Jules for task [13625971885252271733](https://jules.google.com/task/13625971885252271733) started by @arran4*


plan.md
arran4, line 1: @jules complete and clean up please

cmd/g2/site.go
```diff
@@ -444,6 +449,121 @@ func getHighestVersionsAndCount(versions []VersionData) (template.HTML, template
...
+func stripEmail(s string) string {
+	start := strings.Index(s, "<")
+	end := strings.Index(s, ">")
+	if start != -1 && end != -1 && start < end {
+		s = s[:start] + s[end+1:]
+	}
+	return strings.TrimSpace(s)
+}
+
+func parseNewsBodyHTML(body string) template.HTML {
```
arran4: @jules I am expecting parsers to be at / can we also get a model for the news object with the tags etc. That we can reassemble into html.

arran4: > [@jules](https://github.com/jules) I am expecting parsers to be at `/`. Can we also get an ast model for the news object with the tags etc. That we can reassemble into html and txt using serialization functions.

news_parser.go
```diff
@@ -0,0 +1,184 @@
...
+	escaped = strings.ReplaceAll(escaped, "\n", "<br>")
+	return template.HTML(escaped)
+}
+
+// ToText returns the plain text body of the NewsItem.
+func (n NewsItem) ToText() string {
+	return n.Body
+}
+
+func parseNewsBodyHTML(body string) template.HTML {
```
arran4: @jules sorry it was the body (not the header) I wanted converted to an AST you can undo the previous conversion, the conversion here, should be a string to news2ast to html

news_parser.go
```diff
@@ -0,0 +1,184 @@
+package g2
+
+import (
+	"html/template"
+	"strings"
+	"time"
+)
+
+// ParseNewsItem parses the content of a Gentoo news file into a NewsItem structure.
+func ParseNewsItem(content string) NewsItem {
```
arran4, line 10: @jules sorry the header should remain in a struct, and the body should be parsed, and provided as field of said struct.

PR #134, opened 2026-03-27
Interactive Conditional Ebuild Dependencies

arran4:
Implements fully interactive ebuild dependency filtering based on USE conditions provided by the user, deep recursive tree expansions, and inverse dependency lookup utilizing the search index subsystem on the frontend to decrease backend build load. Includes robust testing for new logic.

---
*PR created automatically by Jules for task [6725190638174737076](https://jules.google.com/task/6725190638174737076) started by @arran4*


ebuild.go
```diff
@@ -862,3 +862,29 @@ func PadVersionTokens(v string) string {
 		return fmt.Sprintf("%010s", s)
 	})
 }
+
+// ExtractPackageNameFromDep strips version, slot, and USE flags from a package string
+func ExtractPackageNameFromDep(dep string) string {
```
arran4: @jules this should be using an AST as we will need to extract and use those other components.

cmd/g2/search_index.go
```diff
@@ -142,6 +154,26 @@ func generateSearchIndex(outDir string, sites []*SiteData) error {
...
+								matches := pkgRegex.FindAllString(d, -1)
+								for _, m := range matches {
+									bdepends = append(bdepends, m)
+								}
+							}
+
+							if d := ver.Ebuild.Vars["PDEPEND"]; d != "" {
+								rawPdepends = d
+								matches := pkgRegex.FindAllString(d, -1)
+								for _, m := range matches {
```
arran4: @jules `S1011: should replace loop with pdepends = append(pdepends, matches...) (staticcheck)`

cmd/g2/sitegen_templates/ebuild_details.html
arran4, line 1: @jules shouldn't some of the paths and data be injected?

PR #135, opened 2026-03-27
Fix file name too long error during site license generation

arran4:
Fix file name too long error during site license generation

This commit fixes an issue where null bytes in the LICENSE field of an ebuild (e.g., from upstream corruption or caching errors) would propagate into the generated license string. This would cause the site generator to fail with a "file name too long" error when attempting to create a directory for the corrupt license name. Null bytes are now sanitized and stripped before parsing the license string.

A test case has also been added to ensure `ParseLicense` correctly strips these null characters.

---
*PR created automatically by Jules for task [11379584557027617972](https://jules.google.com/task/11379584557027617972) started by @arran4*


arran4:
@jules this is an incorrect solution and diagnosis of the problem

arran4:
@jules the provided solution is incorrect and the provided update was nil. Can you please use sanity checks to catch long filenames, or anything else you can diagnose as the issue when you generate a site using https://github.com/gentoo/api-gentoo-org/blob/master/files/overlays/repositories.xml Please create sanity and validity checks. 

The fact we were getting nulls or other non ascii characters seems to be a major flaw in our parser can you please determine what is causing that?

PR #138, opened 2026-03-27
Add license sanity lint rule

arran4:
Add a new linting rule `LicenseSanityLintRule` to `lints/ebuild` to check if a `LICENSE` string contains full-text rather than short license identifiers. Tests were added to ensure it correctly catches long text, punctuation-heavy text, or text with too many stop words while properly ignoring normal, valid identifiers.

---
*PR created automatically by Jules for task [223784879772648724](https://jules.google.com/task/223784879772648724) started by @arran4*


arran4:
@jules can you confirm this works with catching the bug that prevents https://github.com/gentoo/api-gentoo-org/blob/master/files/overlays/repositories.xml from executing successfully and what causes it to pick up a lot of bad licenses?

PR #144, opened 2026-03-27
chore: move test-repo directory to testdata

arran4:
Moved the `test-repo/` directory to `testdata/test-repo/` as per the issue request.
No code changes were needed because the tests and codebase do not reference `test-repo/` via hardcoded paths; rather the string "test-repo" is only used as a string literal (e.g. RepoName) inside abstract struct instantiations.

---
*PR created automatically by Jules for task [13873293953809176981](https://jules.google.com/task/13873293953809176981) started by @arran4*


arran4:
@jules I am expecting this to cause some code modifications to change where it's pointing, if not then these were redundant?

arran4:
@jules if they are redundant then they should be removed.

PR #146, opened 2026-03-27
feat: Implement concurrent parsing and shallow/zip downloads

arran4:
This change speeds up repository parsing by processing them in parallel using an `errgroup` worker pool. To minimize bandwidth and local footprint, Git clone commands have been updated to use `--depth 1` to ignore history. Furthermore, a `-use-zip` toggle was added to directly download zip archives for hosts that support it (such as GitHub), safely extracting and stripping root directories. All shared state modifications are protected via `sync.Mutex`.

---
*PR created automatically by Jules for task [2502604992709108841](https://jules.google.com/task/2502604992709108841) started by @arran4*


cmd/g2/site.go
```diff
@@ -2791,45 +2768,108 @@ func (cfg *MainArgConfig) cmdSiteRemote(repositoriesFile string, outDir string,
 	defer func() { _ = os.RemoveAll(tmpDir) }()
 
 	var allSites []*SiteData
+	var sitesMu sync.Mutex
+
+	eg := new(errgroup.Group)
+	eg.SetLimit(runtime.GOMAXPROCS(0))
 
 	for _, repo := range repos.Repositories {
```
arran4, line 2777: @jules we can't do all of them at once, we should have a sane restriction which can be adjusted by a flag.

arran4:
@jules please push changes

cmd/g2/site.go
```diff
@@ -2791,45 +2769,107 @@ func (cfg *MainArgConfig) cmdSiteRemote(repositoriesFile string, outDir string,
...
-		repoPath := filepath.Join(tmpDir, repo.Name)
-		// Try to shallow clone
-		cmd := exec.Command("git", "clone", gitUrl, repoPath)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-		if err := cmd.Run(); err != nil {
-			log.Printf("Failed to clone %s: %v", repo.Name, err)
-			continue
-		}
+			if useZip {
```
arran4, line 2798: @jules make this a function and have an unwrappable error which dicates "fall back to git" and populate it accordingly.

Ensure that we support a registry like system for support of different providers, such as github.com with a basic eval like system (perhaps something like https://pkg.go.dev/github.com/arran4/go-evaluator ) for detecting which backend to use. Git should be fallback but also an option

cmd/g2/site.go
```diff
@@ -2747,7 +2662,70 @@ func renderPage(path string, tmpl *template.Template, name string, data map[stri
 	return nil
 }
 
-func (cfg *MainArgConfig) cmdSiteRemote(repositoriesFile string, outDir string, recentDuration time.Duration, recentDurationStr string, fastGit bool) error {
+func extractZip(zipPath string, destDir string) error {
```
arran4, line 2665: @jules site is getting a bit big this is a util function clearly. If the zipsize isn't too big (compared to memory vs concurrent agents) see if we can do the whole thing in memory rather than extracting.

arran4: > [@jules](https://github.com/jules) site is getting a bit big this is a util function clearly. If the zipsize isn't too big (compared to memory vs concurrent agents) see if we can do the whole thing in memory rather than extracting.

cmd/g2/site.go
arran4: > [@jules](https://github.com/jules) make this a function and have an unwrappable error which dicates "fall back to git" and populate it accordingly.
> 
> Ensure that we support a registry like system for support of different providers, such as github.com with a basic eval like system (perhaps something like https://pkg.go.dev/github.com/arran4/go-evaluator ) for detecting which backend to use. Git should be fallback but also an option

PR #150, opened 2026-03-27
Fix category validation messages and add lint rules

arran4:
Optimize category sanity checks by pre-detecting during parseRepo

- Adjusts log levels for missing categories:
  - Prints a Note if the category isn't in the main Gentoo categories list.
  - Prints a Warning if the category isn't in the repository's `profiles/categories` but is in the main list.
  - Prints an Error if the category isn't in either list.
- Prepends `Note:`, `Warning:`, and `Error:` log messages accurately based on these conditions.
- Limits log spam by placing the `log.Printf` validations strictly inside the check where the category directory has actual ebuild packages (`if len(catData.Packages) > 0`). This prevents false positive complaints about standard non-category directories like `profiles` or `metadata`.
- Optimizes linter performance by directly appending the `Note`, `Warning`, and `Error` messages to the `pkgData.LintWarnings` array during the repository parsing phase, fulfilling the feature request without the runtime cost of executing a dedicated lint rule for each package.

---
*PR created automatically by Jules for task [14785130638873430775](https://jules.google.com/task/14785130638873430775) started by @arran4*


cmd/g2/site.go
```diff
@@ -1075,17 +1079,30 @@ func parseRepo(sysFS fs.FS, repoDir string, defaultTitle string, fastGit bool, r
 
 			pkgData.LintWarnings = lints.PerformLinting(repoDir, &g2PkgData)
 
+			if len(mainCats) > 0 && !inMain {
+				pkgData.LintWarnings = append(pkgData.LintWarnings, fmt.Sprintf("Note: category '%s' is not in the main gentoo categories list", name))
```
arran4: @jules for all the lint warning/errors/notes; we need to add the actual lint rule too the code base too, just pre-calculated here so it doesn't need to run. Can you figure out how.

lints/ebuild/category_sanity.go
arran4, line 1: @jules this need to actually work when run on it's own

arran4: @jules thanks

PR #151, opened 2026-03-27
Fix ebuild parsing of subshell function bodies and quoted array elements

arran4:
The ebuild parser was failing on two main syntax edge-cases frequently found in Gentoo ebuilds:
1. `expected '{' for function body`: Sometimes ebuilds use parentheses for function bodies (`src_prepare() ( ... )`) instead of curly braces.
2. `unterminated array EOF`: Parentheses nested within quoted strings inside array assignments were prematurely decrementing or incrementing the array depth tracking loop, resulting in premature parsing conclusions.

This change fixes both issues:
- `skipFunctionBody` now registers whether `{` or `(` was used, matching it accordingly. If `(` is used, it emits a descriptive warning and treats it as a subshell body without erroring.
- `consumeArray` now ignores any matching parenthesis while iterating through either double quotes `"` or single quotes `'`.
- Both cases are backed up by minimal integration tests.

---
*PR created automatically by Jules for task [12885638716226385325](https://jules.google.com/task/12885638716226385325) started by @arran4*


ebuild_parser.go
```diff
@@ -381,8 +406,17 @@ func (p *EbuildParser) skipFunctionBody() error {
 	if err != nil {
 		return err
 	}
+
+	opener := '{'
+	closer := '}'
 	if r != '{' {
-		return fmt.Errorf("%w: expected '{' for function body", ErrSyntaxError)
+		if r == '(' {
+			log.Printf("Warning: parsing ebuild variables: expected '{' for function body but found '(', treating as subshell body")
```
arran4: @jules we need a lint rule for this, and to aggergate these for the lint rule so it doesn't have to recompute it

arran4: @jules thanks

lints/ebuild/subshell_function.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (l *SubshellFunctionLintRule) Lint(repoDir string, pkgData *g2.PackageData) []string {
+	var results []string
+
+	for _, version := range pkgData.Versions {
+		if version.Ebuild == nil {
+			continue
+		}
+
+		for _, warning := range version.Ebuild.ParseWarnings {
```
arran4, line 33: @jules does this mean we need to move the AST into the package data

PR #172, opened 2026-03-28
Fetch repositories concurrently and add Github/GitLab zip registries

arran4:
This commit implements concurrent repository parsing using `golang.org/x/sync/errgroup` and a `sync.Mutex` in `cmd/g2/site.go` for the `site generate` command and in `cmd/g2/site_serve.go` for the `site serve` command. It also replaces the default git clone with a new `FetchRepo` function, which introduces a `ZipUrlRegistry` to fetch zip archives for GitHub and GitLab when the `-use-zip` toggle is present. If the zip download fails or is unsupported, it falls back to a shallow git clone (`--depth 1`) to skip downloading history.

---
*PR created automatically by Jules for task [16929915709110628093](https://jules.google.com/task/16929915709110628093) started by @arran4*

### DIFF
--- a/cmd/g2/fetch.go
+++ b/cmd/g2/fetch.go
@@ -141,13 +141,18 @@ func downloadAndExtractZip(ctx context.Context, zipUrl string, destDir string) e
 
 		out, err := os.Create(targetPath)
 		if err != nil {
-			rc.Close()
+			_ = rc.Close()
 			return err
 		}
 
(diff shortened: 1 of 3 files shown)


arran4:
After big refactor

arran4:
@jules add more code repository hosts and detect gitlab self hosting IF we can provide gitlab at all

PR #188, opened 2026-03-30
feat: unlimit errgroup concurrency for remote fetching

arran4:
Removes the GOMAXPROCS concurrency limit from the errgroup instances handling repository checkouts in `cmdSiteRemote` and `cmdSiteServe`. Since checkouts are network and I/O bound, restricting them to CPU core count artificially degrades parallel processing performance, forcing sequential behavior on low-core machines.

---
*PR created automatically by Jules for task [987076464177461171](https://jules.google.com/task/987076464177461171) started by @arran4*

### DIFF
--- a/cmd/g2/site.go
+++ b/cmd/g2/site.go
@@ -14,7 +14,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"runtime"
 	"sort"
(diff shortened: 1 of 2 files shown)


arran4:
@jules it's not that the limit should be removed it's that we need a lower limit

PR #206, opened 2026-03-31
feat(cli): add `ebuild diff-json` subcommand

arran4:
Add a command to compare variables retrieved using `ParseFull` and `ParseVariables` to find and pinpoint specific gaps in parser feature parity. Also added `txtar` end to end tests with json snapshots.

---
*PR created automatically by Jules for task [11740609299125488548](https://jules.google.com/task/11740609299125488548) started by @arran4*

### DIFF
--- a/ebuild_diff_txtar_test.go
+++ b/ebuild_diff_txtar_test.go
@@ -1,0 +1,102 @@
+package g2_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"os/exec"
+	"strings"
+	"testing"
+	"os"
+	"path/filepath"
+
+	"golang.org/x/tools/txtar"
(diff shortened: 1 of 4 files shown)


ebuild_diff_txtar_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"os/exec"
+	"strings"
+	"testing"
+	"os"
+	"path/filepath"
+
+	"golang.org/x/tools/txtar"
+)
+
+var diffEbuildsArchive = []byte(`
```
arran4, line 15: @jule this should be an embedded file, it makes no sense to use txtar in this sense, especially if we don't have the expected value here

PR #224, opened 2026-04-02
Implement lazy file content loading via weak pointer

arran4:
Implement lazy file content loading via weak pointer

Added `FileContent` struct inside `site.go` that caches loaded file
contents using Go 1.24 `weak.Pointer[[]byte]`. Provides a `Data()`
method that checks if the weak pointer currently has a value, and
if not uses a generator `func() (io.ReadCloser, error)` to fetch
the content, read it with `io.ReadAll`, and store it back into the
weak pointer. This fulfills the user request to cache file content
in memory without keeping a strong reference indefinitely, avoiding
exhausting memory.

---
*PR created automatically by Jules for task [3577182505465313962](https://jules.google.com/task/3577182505465313962) started by @arran4*


site.go
```diff
@@ -36,28 +37,32 @@ type FileData struct {
...
-		return *ptr, nil
+		return ptr, nil
 	}
 
 	rc, err := fc.generate()
 	if err != nil {
 		return nil, err
 	}
-	defer rc.Close()
+	defer func() { _ = rc.Close() }()
```
arran4: @jules give file content an io.Closer that we pass this too or wrap so we can change the lambda

site.go
```diff
@@ -30,6 +36,35 @@ type FileData struct {
 	RawURL string
 }
 
+type FileContent struct {
```
arran4: @jules give this a better name and implement a non weak-pointer version which is memory backed, and also can you implement a lazy loaded version of tha tone too. This one is technically lazy loadable. Make sure that the users have some way of switching between lazy and not

site.go
```diff
@@ -30,6 +36,146 @@ type FileData struct {
 	RawURL string
 }
 
+type FileContent interface {
```
arran4: @jules these should be in it's own utils package

utils/file_content.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (fc *WeakFileContent) SetGenerator(generate func() (io.ReadCloser, error)) {
+	fc.mu.Lock()
+	defer fc.mu.Unlock()
+	fc.generate = generate
+	fc.data = weak.Pointer[[]byte]{}
+}
+
+type LazyFileContent struct {
```
arran4: @jules is there anyway we can make this more flexible as lazy is more a property than a specific way of storing the data while the other two about storage, so it should be composible in some way.. I am thinking like or better (different) than github.com/arran4/go-be-lazy

arran4: @jules using https://arran4.github.io/blog/post/2026/001-type-switched-variadic-system/ like system like github.com/arran4/go-be-lazy can we make fewer structures and reduce code / duplication?

utils/file_content.go
```diff
@@ -0,0 +1,161 @@
+package utils
+
+import (
+	"io"
+	"sync"
+	"weak"
+)
+
+type FileContent interface {
```
arran4, line 9: @jules Stringer too so it can be used directly in templates. (Or error stringer..)

arran4: @jules can we come up with some other way of reporting the issue that doesn't potentially involve providing raw errors to the consumer of the template data?

utils/file_content.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+type defaultFileContent struct {
+	mu       sync.Mutex
+	store    BytesStore
+	lazy     bool
+	generate func() (io.ReadCloser, error)
+}
+
+func NewFileContent(generate func() (io.ReadCloser, error), opts ...any) FileContent {
```
arran4: @jules TBH generator is optional as the might provide the content directly in many forms, io.reader, string, bytes, etc.

arran4:
@jules I submit for your consideration https://github.com/arran4/go-dynamic-content/releases/tag/v0.0.1

PR #228, opened 2026-04-02
fix: memory leak optimistic optimization with weak.Pointer

arran4:
Replaced direct `map[string]string` memory retention with `weak.Pointer` backed `*FileContent` structs, addressing the persistent OOM leak during overlay generation across large/multiple gentoo profile directories.

---
*PR created automatically by Jules for task [1075446145525523206](https://jules.google.com/task/1075446145525523206) started by @arran4*


arran4:
@jules we need to consider the situation were we streamed the data and can't re-reteve it

cmd/g2/site.go
```diff
@@ -78,6 +80,17 @@ func (f *FileContent) Get() ([]byte, error) {
 		return nil, err
 	}
 
+	if rc, ok := rc.(*os.File); !ok || rc == nil {
```
arran4: @jules this is a lot, can we make it an interface and have different implementations of which FileContent is one, but also needs to be renamed to match the new system, also take into consideration mmap

cmd/g2/site.go
```diff
@@ -52,14 +53,56 @@ type ProfileDescEntry struct {
 
 type SourceURL string
 
+type FileContent struct {
```
arran4: @jules we need a none-weakpointer memory based version, and a lazy loaded version of this

cmd/g2/site.go
```diff
@@ -3289,7 +3381,12 @@ func generateRepoPages(outDir string, tmpl *template.Template, sites []*SiteData
 					return fmt.Errorf("rendering page: %w", err)
 				}
 
-				for fName, fContent := range p.Files {
+				for fName, fc := range p.Files {
+					b, err := fc.Get()
```
arran4: @jules let's make it a Stringer too so we can use it directly in templates

cmd/g2/site.go
```diff
@@ -1622,11 +1738,22 @@ func parseProfilesDirFS(sysFS fs.FS, repoDir string, entries []ProfileDescEntry)
...
 		for _, fname := range fileNames {
-			b, err := fs.ReadFile(sysFS, path2.Join(path, fname))
-			if err == nil {
-				pData.Files[fname] = string(b)
+			filePath := path2.Join(path, fname)
+			info, err := fs.Stat(sysFS, filePath)
+			if err == nil && !info.IsDir() {
+				fc := &WeakFileContent{
+					loader: func() (io.ReadCloser, error) {
+						f, err := sysFS.Open(filePath)
```
arran4: @jules this only works if we are not deleting the repo, can you add that condition. Also can you apply this logic elsewhere such as ebuilds and anywhere else we are keeping a copy of the file in memory?

cmd/g2/site.go
```diff
@@ -52,14 +53,129 @@ type ProfileDescEntry struct {
...
+	b, err := io.ReadAll(rc)
+	if err != nil {
+		return nil, err
+	}
+
+	f.strongPtr = b
+	return &f.strongPtr, nil
+}
+
+type MmapFileContent struct {
```
arran4: @jules I feel as though this is redundant

file_content.go
arran4, line 1: @jules can we variadict the weak and lazy like github.com/arran4/go-be-lazy ie https://arran4.github.io/blog/post/2026/001-type-switched-variadic-system/

arran4:
@jules I submit for your consideration https://github.com/arran4/go-dynamic-content/releases/tag/v0.0.1

arran4:
Recreated

PR #229, opened 2026-04-02
Refactor template view directory structure to use nested folders

arran4:
Reorganize `templates/views/` directory by moving templates with an underscore in their name into nested directories (e.g., `repo_package.html` -> `repo/package.html`). Updates `embed.go`, `template_loader.go`, and `renderPage` calls across the project to use the new nested names.

---
*PR created automatically by Jules for task [17846178877316854572](https://jules.google.com/task/17846178877316854572) started by @arran4*


arran4:
@jules okay we will have to be a bit smarter about it.. Because we have a global package file and a repo package file. The common parts should be extracted into partials, and then we can have 2 entry points in appropriate locations with any location specific modifications in the right part. That way we can ensure that the site is logical

arran4:
@jules that's beginning to look good can you continue? Remember to use define/template/etc to allow for multiple versions of the page for different locations. Reflecting how they are outputed is very appealing

PR #234, opened 2026-04-02
🔒 Fix Zip Slip vulnerability in zip extraction

arran4:
🎯 **What:** The vulnerability fixed
A Zip Slip (path traversal) vulnerability in the zip archive extraction process within `cmd/g2/fetch.go`.

⚠️ **Risk:** The potential impact if left unfixed
A malicious zip file containing path traversal characters (`../`) could write files outside of the target extraction directory. This could lead to arbitrary file overwrite, which could be leveraged to execute remote code or compromise the system executing the tool.

🛡️ **Solution:** How the fix addresses the vulnerability
The fix implements `filepath.IsLocal` to guarantee the requested relative path explicitly points to a safe, confined path. It also defensively accounts for trailing zip headers that evaluate to an empty path after structural processing to ensure healthy zip directories extract without functional regression. A direct unit test `TestDownloadAndExtractZip_ZipSlip` was added to assert failure paths against explicit payload testing.

---
*PR created automatically by Jules for task [3671366823300211829](https://jules.google.com/task/3671366823300211829) started by @arran4*

### DIFF
--- a/cmd/g2/fetch_test.go
+++ b/cmd/g2/fetch_test.go
@@ -8,6 +8,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"golang.org/x/tools/txtar"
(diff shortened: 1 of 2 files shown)


cmd/g2/fetch_test.go
```diff
@@ -90,3 +91,48 @@ package main
...
+		t.Fatal(err)
+	}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/zip")
+		_, _ = w.Write(buf.Bytes())
+	}))
+	defer ts.Close()
+
+	destDir, err := os.MkdirTemp("", "g2-fetch-zipslip-test-*")
```
arran4: @jules https://github.com/arran4/blog/blob/main/content/post/2026/007-Go-FSs-Everywhere/index.md

arran4: @jules it's more about the use of memory fs rather than disk, over txtar

arran4: @jules refactor

PR #237, opened 2026-04-02
🔒 Fix XSS Vulnerability in Template Execution

arran4:
🎯 **What:** This PR fixes a Cross-Site Scripting (XSS) vulnerability stemming from the explicit bypass of Go's `html/template` default contextual escaping.

⚠️ **Risk:** The previous implementation rendered a sub-template into a string, casted the string to `template.HTML` (marking it safe), and then injected it into `layout.html`. Because `template.HTML` completely disabled Go's robust built-in HTML contextual escaping for the wrapper, there was an implicit risk of XSS if any dynamic content was rendered in a way the html engine could not trace. Code analysis tools like `gosec` also routinely flag this (`G203`) as an untrusted HTML injection vulnerability.

🛡️ **Solution:** The monolithic `layout.html` file has been cleanly split around the target content block into `layout_header.html` and `layout_footer.html`. `site_serve.go` and `site.go` have been refactored to sequentially execute the header, the page template, and then the footer into the target stream. This eliminates the need for `template.HTML()` completely and correctly restores the `html/template` engine's ability to maintain context and escape user input safely. Tested successfully offline via Playwright and standard Go tests.

---
*PR created automatically by Jules for task [9226240849349147827](https://jules.google.com/task/9226240849349147827) started by @arran4*


cmd/g2/site.go
```diff
@@ -3793,33 +3705,23 @@ func generateSite(outDir string, sites []*SiteData, recentDuration time.Duration
...
 
 	f, err := os.Create(path)
 	if err != nil {
 		return fmt.Errorf("creating file %s: %w", path, err)
 	}
 	defer func() { _ = f.Close() }()
 
-	if err := tmpl.ExecuteTemplate(f, "layout.html", layoutData); err != nil {
-		return fmt.Errorf("executing layout template for %s: %w", path, err)
+	if err := tmpl.ExecuteTemplate(f, "layout_header.html", data); err != nil {
```
arran4, line 3715: @jules isn't better to make these partials and import them?

arran4: @jules craft a prompt for me to fix this and other redundant parts of the template.

PR #259, opened 2026-04-04
feat: add free memory logging

arran4:
Implemented memory usage logging alongside disk usage logging.

---
*PR created automatically by Jules for task [519067019425629367](https://jules.google.com/task/519067019425629367) started by @arran4*


cmd/g2/site.go
arran4, line 1: @jules what about used mem?

arran4: @jules at the end and every so many repospitories we need a summary table of memory usage and item count for each data set aggregated.

PR #266, opened 2026-04-05
test: Add comprehensive tests for ResolveVariables

arran4:
Added tests to verify the behavior of variable resolution across various complex scenarios, capturing current behavior and providing a safety net for any future refactoring.

---
*PR created automatically by Jules for task [16340413005063095212](https://jules.google.com/task/16340413005063095212) started by @arran4*

### DIFF
--- a/ebuild_test.go
+++ b/ebuild_test.go
@@ -391,6 +391,9 @@ func TestResolveVariables(t *testing.T) {
 		},
 		{
 			name: "Bash conditionals are treated as literal strings",
+			// TODO: ResolveVariables currently does not evaluate bash conditionals (like if/else),
+			// it treats them as literal strings and only performs variable substitution.
+			// This is a known limitation that needs to be addressed for full bash compliance.
 			// Proves that bash conditional logic is NOT evaluated, only variables within them are substituted.
 			text: "if [[ $A == \"foo\" ]]; then echo $B; fi",
 			variables: map[string]string{
--- a/ebuild_test.go
+++ b/ebuild_test.go
@@ -234,6 +234,164 @@ func TestBlackbox(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


ebuild_test.go
arran4, line 1: @jules how do we test sequences usch as:
```
a = 5 
a = a * 10
```

Also how about errors such as:
```
a =a 
```

arran4: @jules don't worry about arithmetic then 
```
a = "a"
a = "a$a"
```

etc

arran4: @jules we need to make sure that variable reference is "as we go" so if `a` is replaced the new one refers to the old one, we need a test to prove that this is what we are doing. Thus in the situation:
```
a = a
```
`a` needs to be injected into the context, and becomes a runtime requirement.

arran4: @jules ensure:
```
a = "a"
b = "$a"
a = "$a$a"
b = "$b"
```
Resolves correctly, which it might not do if the references aren't created and referenced line by line.

arran4: @jules what are the consequences of this distinction. Or is this for the final resolved values?

arran4: @jules won't this be an issue with the complexity of ebuilds?

arran4: @Jules typically loops are not used for metadata but conditions certianly are. Do we support the resolution of that in this, if so can we add a couple to the test suite. - Actually either way to cememt that it does or doesnt' happen

arran4: @jules yes please I think we can agree that that is an oversight that will need to be addressed to given the context of what this is showing?

arran4: @jules add a todo to it,and give me a prompt to run to resolve this issue

PR #268, opened 2026-04-06
Log heap objects when concurrency is 1

arran4:
Log object stats like `HeapObjects` and `Alloc` after every repository is successfully parsed when `-concurrency=1` is provided.

---
*PR created automatically by Jules for task [15036996173284028646](https://jules.google.com/task/15036996173284028646) started by @arran4*

### DIFF
--- a/cmd/g2/site.go
+++ b/cmd/g2/site.go
@@ -4254,6 +4254,36 @@ func (cfg *MainArgConfig) cmdSiteRemote(repositoriesFile string, outDir string, 
 			}
 
 			if concurrency == 1 {
+				numCategories := len(siteData.Categories)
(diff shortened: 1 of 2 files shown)


cmd/g2/site.go
```diff
@@ -4253,6 +4253,16 @@ func (cfg *MainArgConfig) cmdSiteRemote(repositoriesFile string, outDir string,
 				)
 			}
 
+			if concurrency == 1 {
```
arran4, line 4256: @jules we should be doing this frequently already, and i was expecting it to report things n number of ebuilds, n nubmer of repos, n number of profiles, n number of licenses and so forth

PR #269, opened 2026-04-06
feat: allow specifying custom temporary directory for site generation

arran4:
Added a `-temp-dir` flag to `overlay site generate` and `overlays site generate` commands to allow users to override the system default temporary directory when cloning repositories.

---
*PR created automatically by Jules for task [15425427180103090151](https://jules.google.com/task/15425427180103090151) started by @arran4*

### DIFF
--- a/cmd/g2/site.go
+++ b/cmd/g2/site.go
@@ -220,7 +220,7 @@ func (cfg *MainArgConfig) cmdOverlay(args []string) error {
 	recentDurOpt := fs.String("recent-duration", "3mo", "Duration to consider an update 'recent' (e.g. 3mo, 14d, 72h)")
 	fastGit := fs.Bool("fast-git-modtime", false, "Use fast (O(1)) but potentially less reliable go-git file log lookup")
 	useZip := fs.Bool("use-zip", false, "Download zip archives instead of git clone when supported")
-	mode := fs.String("mode", "persistent", "Clone mode: 'persistent' or 'temp'")
+	workMode := fs.String("work-mode", "persistent", "Work mode: 'persistent' or 'temp'")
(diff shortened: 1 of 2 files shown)


cmd/g2/site.go
```diff
@@ -213,6 +213,7 @@ func (cfg *MainArgConfig) cmdOverlay(args []string) error {
 	fastGit := fs.Bool("fast-git-modtime", false, "Use fast (O(1)) but potentially less reliable go-git file log lookup")
 	useZip := fs.Bool("use-zip", false, "Download zip archives instead of git clone when supported")
 	persistentDir := fs.String("persistent-dir", "", "Directory to persistently store checked out repositories instead of a temporary directory")
+	tempDir := fs.String("temp-dir", "", "Directory to use for temporary files instead of the default")
```
arran4, line 225: @jules is "temp" the best name, as it's more of a cache, we could use xdg cache...

arran4: @jules we will need different modes, persistant and temp. I would also like to switch the default mode to persistent. We also need a good fall back if XDP cache is not available (such as running as root, or running on CI.)

arran4: @jules it should be work-mode or persistance-mode or something not mode as there could be more than that.

cmd/g2/site.go
```diff
@@ -313,7 +323,7 @@ func (cfg *MainArgConfig) cmdOverlay(args []string) error {
 				defer cancel()
 
 				t0 := time.Now()
-				if err := FetchRepo(ctx, task.Location, tmpDir, *useZip, *persistentDir != "", 0); err != nil {
+				if err := FetchRepo(ctx, task.Location, tmpDir, *useZip, *workMode == "persistent", 0); err != nil {
```
arran4: @jules we might support more modes let's not convert them to a bool like this but rather code paths or switches etc

arran4:
Redo conflcits

PR #273, opened 2026-04-08
Remove duplicate assignment warning in ebuild_parser.go

arran4:
Fixes the duplicate assignment warning in `ebuild_parser.go` which happens whenever a variable is assigned twice. Because variables can be reassigned during evaluation and it is valid.

---
*PR created automatically by Jules for task [13918315615769844765](https://jules.google.com/task/13918315615769844765) started by @arran4*


arran4:
@jules can you verify that we are processing this correctly still?

PR #278, opened 2026-04-08
Fix memory leak using go-weak-content

arran4:
Fix massive memory leak by using go-weak-content for Ebuild RawText and Functions

---
*PR created automatically by Jules for task [10809439597152313280](https://jules.google.com/task/10809439597152313280) started by @arran4*

### DIFF
--- a/cmd/g2/ebuild.go
+++ b/cmd/g2/ebuild.go
@@ -378,13 +378,14 @@ func (cfg *CmdEbuildArgConfig) cmdEbuildExplain(args []string, opts ...any) erro
 	_, _ = fmt.Fprintln(out)
 
 	_, _ = fmt.Fprintf(out, "=== Phases Overridden ===\n")
-	if len(ebuild.Functions) > 0 {
-		var funcs []string
-		for f := range ebuild.Functions {
-			funcs = append(funcs, f)
+	funcs, _ := ebuild.FunctionsContent.Data()
+	if funcs != nil && len(*funcs) > 0 {
+		var funcNames []string
+		for f := range *funcs {
+			funcNames = append(funcNames, f)
 		}
-		sort.Strings(funcs)
-		for _, f := range funcs {
+		sort.Strings(funcNames)
+		for _, f := range funcNames {
 			_, _ = fmt.Fprintln(out, f)
 		}
 	} else {
--- a/cmd/g2/eclass.go
+++ b/cmd/g2/eclass.go
@@ -163,15 +163,18 @@ func (cfg *CmdEclassArgConfig) cmdEclassExplain(args []string) error {
 	}
 
 	fmt.Println("\nFunctions:")
-	for k, v := range ebuild.Functions {
-		fmt.Printf("  %s() {\n", k)
-		lines := strings.Split(v.Value, "\n")
-		for _, line := range lines {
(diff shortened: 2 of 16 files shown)


arran4:
@jules can you bench mark this?

arran4:
@jules how about against https://github.com/gentoo/api-gentoo-org/blob/master/files/overlays/repositories.xml ?

arran4:
@jules is that because of lazy evaluation, which means we have to run the whole thing. It might be slower as page generation is done serialy right?

arran4:
@jules can you bench mark it against https://github.com/gentoo/api-gentoo-org/blob/master/files/overlays/repositories.xml so I can see what the impact is. the expected runtime for each is 1 hour 30 minutes, least on github actions it is.

arran4:
@Jules can you emulate memory constrained and find more suggestions as what we should do. Using realworld respositories.xml as that's what I'm running it against

arran4:
> > @jules can you emulate memory constrained and find more suggestions as what we should do. Using realworld respositories.xml as that's what I'm running it against
> 
> I emulated a memory-constrained environment using `GOMEMLIMIT=250MiB` and ran it against the real-world `repositories.xml` for several iterations, also generating heap profiles.
> 
> Here are some concrete suggestions for further memory optimization (beyond just `go-weak-content`) that surfaced from the profiling sessions in the real-world dataset:
> 
>     1. **Shared String Caching / Interning:** We parse strings like "DEPEND", "RDEPEND", architectures (like `amd64`, `x86`), and category names (`dev-python`, `sys-apps`) across hundreds of thousands of files. These identical strings take up completely separate memory allocations. Introducing a simple string interning cache (`map[string]string`) or utilizing Go 1.24+ `unique` package could drastically slash the `Vars` map allocation footprints.
> 
>     2. **Abstract Syntax Tree (AST) Simplification:** The `g2` ebuild parser creates an AST with maps and deep slices for each function body. While `weak-content` mitigates the long-term cost, the transient spike per repository remains high. By parsing `Functions` strictly on an as-needed basis (i.e. only tokenizing and verifying syntax without converting to `map[string]AST` structs unless actively requested for display) or representing them as just string byte-slices, we avoid thousands of tiny node object allocations.
> 
>     3. **Optimize `ResolveVariables` Allocation:** The benchmark profiles show that `ResolveVariables` generates high GC pressure due to repeated loop operations and regex compilations (even with `ReplaceAllStringFunc`). Implementing faster-path byte scanning instead of regexp operations and pre-allocating known limits would drastically cut short-lived memory limits.
> 
>     4. **Sequential Manifest Extraction:** Rather than building all `ManifestData` structs for the entire repo at once, streaming this lazily using an iterator when the respective templates are rendered would further avoid keeping large data slices in memory.
> 
> 
> The current `go-weak-content` patch successfully prevents the application from immediately hard-crashing under constrained environments by dropping those objects when `GOMEMLIMIT` forces aggressive garbage collection cycles, but implementing string interning would be the best next step to lower the overall peak memory ceiling across workers!

@jules

Definitely do 1.

Number 2: AST I intend to do more with the data, but you're right if we can introduce a mode for lesser compilations, or even do a degree of lazy evaluations it would drastically help

Number 3: I have another PR for this, please give me a prompt here to run in the future. 

Number 4: Yes please.

Thanks

PR #279, opened 2026-04-08
Refactor ebuild variable parser to evaluate bash conditionals with mvdan.cc/sh

arran4:
Refactored variable resolution logic using mvdan.cc/sh/v3 to correctly evaluate bash conditionals instead of parsing them as literal strings. Added restricted ExecHandler logic for safety and ensured indirect/chained variable resolution works natively as bash handles it.

---
*PR created automatically by Jules for task [9041384759997853771](https://jules.google.com/task/9041384759997853771) started by @arran4*


vars-1.0.tar.gz
arran4, line 1: @jules what is this?

resolver.go
```diff
@@ -0,0 +1,72 @@
...
+					if len(args) > 0 && args[0] == "echo" {
+						return next(ctx, args)
+					}
+					// Deny all other external commands
+					return fmt.Errorf("external command execution denied: %s", args[0])
+				}
+			}),
+		)
+		if err == nil {
+			err = runner.Run(context.Background(), file)
```
arran4: @jules resolveBash should take a context and we should be able to configure individual as well as global time limits using the context and variadict args

ebuild_test.go
arran4, line 1: @jules lots of rambling here can you resolve that?

vars-1.0.tar.gz
arran4, line 1: @jules please clean this file up

arran4:
@jules please add another test that would fail under the former version but pass in this one?

PR #290, opened 2026-04-11
Optimize search index payload size

arran4:
Reduce size of JSON files generated by search index by omitting empty fields.

---
*PR created automatically by Jules for task [16150749914567074223](https://jules.google.com/task/16150749914567074223) started by @arran4*

### DIFF
--- a/cmd/g2/search_index.go
+++ b/cmd/g2/search_index.go
@@ -17,37 +17,37 @@ import (
 
 type SearchDocument struct {
 	ID              int      `json:"id"`
-	Overlay         string   `json:"overlay"`
-	Category        string   `json:"category"`
-	Package         string   `json:"package"`
-	FullName        string   `json:"full_name"`
-	Version         string   `json:"version"`
-	VersionSortKey  string   `json:"version_sort_key"`
-	Description     string   `json:"description"`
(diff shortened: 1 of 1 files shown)


arran4:
@jules can you propose algorithmic changes to the index to reduce load size, perhaps instead of fixing them to a file size using a smaller partition? Please test suggestions and provide differences and benchmarks

arran4:
@jules let's move to a precomuted inverted index please. If we combine it with a REST style path lookup does that also help?

arran4:
> > @jules let's move to a precomuted inverted index please. If we combine it with a REST style path lookup does that also help?
> 
> Yes, using an inverted index combined with a REST style path lookup would drastically improve performance and reduce the amount of data loaded by the client. It avoids downloading a giant monolithic index or large chunks.
> 
> Since this would require significant architectural changes to `search_index.go`, the chunking logic, and rewriting `search.js` / `search_ui.js`, it makes sense to address that in a completely separate task/PR. The `omitempty` change in _this_ PR provides an immediate, safe ~40% size reduction on the current flat JSON architecture. I recommend merging this quick win first, and then we can tackle the precomputed inverted index implementation separately as it is a major rewrite.

Please give me a detailed prompt

PR #357, opened 2026-07-23
Filter full text licenses and separate unsupported licenses

arran4:
This PR filters out full-text licenses from the license database and separates licenses into supported (having a file in an overlay) and unsupported (not having a file in any overlay), generating a new `/unsupported-licenses/` page for the latter.

---
*PR created automatically by Jules for task [12987271553151114316](https://jules.google.com/task/12987271553151114316) started by @arran4*

### DIFF
--- a/cmd/g2/site_serve.go
+++ b/cmd/g2/site_serve.go
@@ -256,7 +256,6 @@ type SiteServer struct {
 	AggCategories          []*AggCategory
 	AggPackages            []*AggPackage
 	AggLicenses            []*AggLicense
-	AggSupportedLicenses   []*AggLicense
 	AggUnsupportedLicenses []*AggLicense
 	AggProjects            []*AggProject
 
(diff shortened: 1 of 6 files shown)


cmd/g2/site_serve.go
```diff
@@ -249,14 +249,16 @@ func isOverlayDir(dir string) bool {
...
-	AggLicenses   []*AggLicense
-	AggProjects   []*AggProject
+	GenInfo                GenerationInfo
+	tmpl                   *template.Template
+	Title                  string
+	Sites                  []*g2.SiteData
+	AggCategories          []*AggCategory
+	AggPackages            []*AggPackage
+	AggLicenses            []*AggLicense
+	AggSupportedLicenses   []*AggLicense
```
arran4: @jules let's filter out unsupportedlicenses so we don't need "supported licenses" also complete the ticket.

arran4: > I've removed `Licenses` from the dashboard context entirely since we now have `SupportedLicenses` and `UnsupportedLicenses`.

@jules doesn't that require a massive refactor to make it consistent and not break anything? I am expecting site implications too. Also I would rather call `SupportedLicenses`just `Licenses` then `UnsuportedLicenses` is it's own that people can check for various reasons

arran4: > > I've removed `Licenses` from the dashboard context entirely since we now have `SupportedLicenses` and `UnsupportedLicenses`.
> 
> [@jules](https://github.com/jules) doesn't that require a massive refactor to make it consistent and not break anything? I am expecting site implications too. Also I would rather call `SupportedLicenses`just `Licenses` then `UnsuportedLicenses` is it's own that people can check for various reasons

PR #381, opened 2026-07-28
feat(cli): add `g2 ebuild upsert` command

arran4:
Implement `g2 ebuild upsert` to safely manage new ebuild versions from stdin.

The `g2 ebuild upsert` subcommand reads ebuild content from standard input, compares it to the highest existing revision of the specified package and version in a directory, and writes a new bumped revision (e.g., `-r1`) only if the actual ebuild content differs. It also supports an `--ignore-comments` flag to prevent trivial bumps when only comments (like generated timestamps) change.

This logic fixes an issue in workflow regeneration tools that typically check for the base version and enter an infinite loop of creation/deletion by automatically incrementing revisions inline.

---
*PR created automatically by Jules for task [9416026059292226307](https://jules.google.com/task/9416026059292226307) started by @arran4*

### DIFF
--- a/cmd/g2/ebuild.go
+++ b/cmd/g2/ebuild.go
@@ -106,6 +106,10 @@ func (cfg *MainArgConfig) cmdEbuild(args []string) error {
 		if err := config.cmdEbuildNextRevision(fs.Args()[1:]); err != nil {
 			return fmt.Errorf("ebuild next-revision: %w", err)
 		}
+	case "upsert":
+		if err := config.cmdEbuildUpsert(fs.Args()[1:]); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/g2/ebuild.go
```diff
@@ -106,6 +106,10 @@ func (cfg *MainArgConfig) cmdEbuild(args []string) error {
 		if err := config.cmdEbuildNextRevision(fs.Args()[1:]); err != nil {
 			return fmt.Errorf("ebuild next-revision: %w", err)
 		}
+	case "upsert":
```
arran4, line 109: @jules isn't this redundant compared to `g2 next-revision` ?

PR #383, opened 2026-07-29
test: add tests and usage for ebuild upsert

arran4:
Adds the previously missing usage string in `g2 ebuild --help` for the `upsert` command and provides comprehensive unit tests for `ebuild upsert` functionality, addressing issue #375 (since the command itself was already introduced).

---
*PR created automatically by Jules for task [17398981520432264606](https://jules.google.com/task/17398981520432264606) started by @arran4*


cmd/g2/ebuild.go
```diff
@@ -39,6 +39,7 @@ func (cfg *MainArgConfig) cmdEbuild(args []string) error {
 		fmt.Printf("\t\t %s \t\t %s\n", "tag", "Ebuild specific tag subcommand supporting version comparisons")
 		fmt.Printf("\t\t %s \t\t %s\n", "check-exists", "Determine if any revision of a given version exists")
 		fmt.Printf("\t\t %s \t\t %s\n", "next-revision", "Generate the next revision name and optionally inspect contents")
+		fmt.Printf("\t\t %s \t\t %s\n", "upsert", "Upsert an ebuild from stdin")
```
arran4: @jules technically it's a revision upsert right?

fix_errcheck.patch
arran4, line 1: @jules please clear out your working files like this *.patch file

cmd/g2/ebuild.go
```diff
@@ -39,6 +39,7 @@ func (cfg *MainArgConfig) cmdEbuild(args []string) error {
 		fmt.Printf("\t\t %s \t\t %s\n", "tag", "Ebuild specific tag subcommand supporting version comparisons")
 		fmt.Printf("\t\t %s \t\t %s\n", "check-exists", "Determine if any revision of a given version exists")
 		fmt.Printf("\t\t %s \t\t %s\n", "next-revision", "Generate the next revision name and optionally inspect contents")
+		fmt.Printf("\t\t %s \t\t %s\n", "upsert", "Upsert an ebuild revision from stdin")
```
arran4: @jules I expect the `upsert` to be more clear. Also not always from stdin we should allow `-` to define stdin, otherwise <filename> is provided.

`upsert` should perhaps be `upsert-revision` OR simply merged with `next-revision` or made into `next-version` where the difference is we specify the version and then if it's pre-existing it does what `next-revision` does.

Perhaps we should add named bumps as well as a version bump like github.com/arran4/git-tag-inc such that:

`g2 ebuild upsert-version revision <filename>` does the same as `next-revision` and obsoletes it?

arran4: @jules did you allow the user to state to be what to bump and to allow the user to alternatively state the exact version of the ebuild from the ebuild contents?

cmd/g2/ebuild_upsert.go
```diff
@@ -82,13 +82,33 @@ func (cfg *CmdEbuildArgConfig) cmdEbuildUpsert(args []string) error {
...
+		versionToUse = *verFlag
+	} else {
+		parsedVars := g2.ParseEbuildVariablesFromReader(bytes.NewReader(inputContent))
+		if parsedVars != nil && parsedVars["PV"] != "" {
+			versionToUse = parsedVars["PV"]
+		}
+	}
+
+	if *dirFlag == "" || *pkgFlag == "" || versionToUse == "" {
+		return fmt.Errorf("usage: g2 ebuild upsert --dir <ebuildDir> --package <pkgName> [--version <version>] [--ignore-comments] [filename|-]")
```
arran4: @jules
```
return fmt.Errorf("usage: g2 ebuild upsert --dir <ebuildDir> --package <pkgName> [--version <version>] [--ignore-comments] [filename|-]")
```
should be
```
return fmt.Errorf("usage: g2 ebuild upsert --dir <ebuildDir> --package <pkgName> [--version <version>] [--ignore-comments] [new version|version bump type] [filename|-]")
```

arran4: > Updated the usage string to include `[new version|version bump type]` as requested.

@jules it's no good just updating the usage text, we need to make sure that statement is true throughout the code base.

cmd/g2/ebuild_upsert.go
```diff
@@ -82,10 +82,22 @@ func (cfg *CmdEbuildArgConfig) cmdEbuildUpsert(args []string) error {
...
 	}
 
+	var newVersionOrBumpType string
 	var inputContent []byte
 	filename := "-"
+
 	if fs.NArg() > 0 {
-		filename = fs.Arg(0)
+		arg0 := fs.Arg(0)
+		if fs.NArg() == 1 {
```
arran4: @jules it looks like we are doing work to make it consistent we can change the args to make the code easier.

Also please clean up your files like the *.diff files

cmd/g2/ebuild_upsert.go
```diff
@@ -154,14 +181,30 @@ func (cfg *CmdEbuildArgConfig) cmdEbuildUpsert(args []string) error {
...
-	if err := os.WriteFile(targetFile, stdinContent, 0644); err != nil {
+	if err := os.WriteFile(targetFile, inputContent, 0644); err != nil {
 		return fmt.Errorf("failed to write ebuild file %s: %w", targetFile, err)
 	}
 
 	fmt.Println(targetFile)
 	return nil
 }
+
+func fileExists(filename string) bool {
```
arran4: @jules func fileExists is unused (unused)

PR #420, opened 2026-08-09
🧪 Add unit test for ParsePkgDescIndexFile

arran4:
🎯 **What:** Added a unit test `TestParsePkgDescIndexFile` in `pkg_desc_index_test.go` for the previously untested `ParsePkgDescIndexFile` public function.
📊 **Coverage:** The test covers both the successful parsing of a temporary `pkg_desc_index` file created using `t.TempDir()` and the error handling for attempting to parse a non-existent file path.
✨ **Result:** Improved test coverage for file parsing functions in the `g2` package, ensuring correct behavior and preventing future regressions.

---
*PR created automatically by Jules for task [12712602649825205754](https://jules.google.com/task/12712602649825205754) started by @arran4*

### DIFF
--- a/pkg_desc_index.go
+++ b/pkg_desc_index.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"sort"
 	"strings"
@@ -73,9 +74,25 @@ func ParsePkgDescIndex(r io.Reader) (*PkgDescIndex, error) {
 	return &index, nil
 }
 
-// ParsePkgDescIndexFile reads and parses a pkg_desc_index file from disk.
-func ParsePkgDescIndexFile(path string) (*PkgDescIndex, error) {
-	file, err := os.Open(path)
+// ParsePkgDescIndexFile reads and parses a pkg_desc_index file.
+// It optionally accepts an fs.FS to read from via type-switched variadic args.
+// If no fs.FS is provided, it falls back to os.Open.
+func ParsePkgDescIndexFile(path string, opts ...any) (*PkgDescIndex, error) {
+	var fsys fs.FS
+	for _, opt := range opts {
+		if v, ok := opt.(fs.FS); ok {
+			fsys = v
+		}
+	}
+
+	var file io.ReadCloser
+	var err error
+	if fsys != nil {
+		file, err = fsys.Open(path)
+	} else {
+		file, err = os.Open(path)
+	}
+
 	if err != nil {
 		return nil, fmt.Errorf("opening file: %w", err)
 	}
--- a/pkg_desc_index.go
+++ b/pkg_desc_index.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"sort"
 	"strings"
@@ -71,6 +72,16 @@ func ParsePkgDescIndex(r io.Reader) (*PkgDescIndex, error) {
 	}
 
 	return &index, nil
+}
+
+// ParsePkgDescIndexFS reads and parses a pkg_desc_index file from an fs.FS.
+func ParsePkgDescIndexFS(fsys fs.FS, path string) (*PkgDescIndex, error) {
+	file, err := fsys.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("opening file: %w", err)
+	}
+	defer func() { _ = file.Close() }()
+	return ParsePkgDescIndex(file)
 }
 
 // ParsePkgDescIndexFile reads and parses a pkg_desc_index file from disk.
--- a/pkg_desc_index.go
+++ b/pkg_desc_index.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"io/fs"
 	"os"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


pkg_desc_index_test.go
```diff
@@ -75,3 +77,44 @@ app-admin/amazon-ec2-init 20101127-r2: Init script to setup Amazon EC2 instance
...
+
+func TestParsePkgDescIndexFile(t *testing.T) {
+	// Test with a valid file
+	tempDir := t.TempDir()
+	filePath := filepath.Join(tempDir, "pkg_desc_index")
+
+	input := `app-admin/aerospike-amc-community 4.0.19-r2 5.0.0: Web UI based monitoring tool for Aerospike Community Edition Server
+app-admin/amazon-ec2-init 20101127-r2: Init script to setup Amazon EC2 instance parameters
+`
+	err := os.WriteFile(filePath, []byte(input), 0644)
```
arran4, line 89: @jules refactor to use mapfs or fs.fs like systems

pkg_desc_index.go
```diff
@@ -73,6 +74,16 @@ func ParsePkgDescIndex(r io.Reader) (*PkgDescIndex, error) {
 	return &index, nil
 }
 
+// ParsePkgDescIndexFS reads and parses a pkg_desc_index file from an fs.FS.
+func ParsePkgDescIndexFS(fsys fs.FS, path string) (*PkgDescIndex, error) {
```
arran4: @jules it looks like you're correctly added this for tests, but it doesn't look like it's actually used in that code base

pkg_desc_index_test.go
arran4: > Added `ParsePkgDescIndexFS` that accepts `fs.FS` in `pkg_desc_index.go` and refactored testing to use `fstest.MapFS` in a new `TestParsePkgDescIndexFS` test.

@jules you have undone this please redo

arran4, line 89: @jules the tests are still doing direct fs

arran4: @jules can we use a fs.FS like implemtnation ?

arran4:
@jules
Please finish the filesystem/testability refactor around `ParsePkgDescIndexFile`.

Review the current state at commit `d5b30a0` before changing anything. The previous iterations have misunderstood the intent, so please follow the requirements below rather than reverting back to `t.TempDir()`.

References for the intended patterns:

* https://arran4.github.io/blog/post/2026/008-memory-fs-for-testing/
* https://arran4.github.io/blog/post/2026/020-optional-dependency-injection-via-type-switched-variadic-args/
* https://arran4.github.io/blog/post/2026/018-txtar-patterns-for-agents-v2/
* Also review the newer relevant guidance:
  https://arran4.github.io/blog/post/2026/033-testing-fs-with-mapfs-mockfs/

The current implementation is only halfway there:

* `cmdPkgDescIndexVerify` now constructs `g2.NewOsCacheFS(*repoDir)` and manually opens `metadata/pkg_desc_index`, then passes the resulting reader to `ParsePkgDescIndex`.
* As a result, the production code no longer actually uses `ParsePkgDescIndexFile`.
* `TestParsePkgDescIndexFile` still uses `t.TempDir()`, `os.WriteFile`, and the real filesystem.
* `ParsePkgDescIndexFile` itself still hard-codes `os.Open`.

That is not the intended result.

Please make the smallest coherent change so that the filesystem abstraction is exercised by both production and tests.

Preferred design:

1. Preserve the existing `ParsePkgDescIndexFile(path string)` calling convention for existing callers by making filesystem injection optional, following the type-switched variadic options pattern used elsewhere in this repository/blog, e.g. conceptually:

   `ParsePkgDescIndexFile(path string, opts ...any)`

2. Allow an `fs.FS` supplied through `opts` to provide the file.

3. With no filesystem option supplied, preserve the existing real-OS behaviour and compatibility of `ParsePkgDescIndexFile(path)`.

4. Do not introduce a large new abstraction merely for this parser. This is read-only I/O, so standard `io/fs` / `fs.FS` is sufficient.

5. Update `cmdPkgDescIndexVerify` so its existing `g2.NewOsCacheFS(*repoDir)` is passed into `ParsePkgDescIndexFile` with the repository-relative path, rather than manually opening the file and bypassing `ParsePkgDescIndexFile`.

   The goal is for the filesystem-aware public path to be genuinely used by production code, not to exist solely for tests.

6. Refactor `TestParsePkgDescIndexFile` to use `testing/fstest.MapFS`.

   The unit test must not use:

   * `t.TempDir()`
   * `os.WriteFile`
   * `os.Create`
   * other real-disk setup

   Put `pkg_desc_index` directly into an in-memory `fstest.MapFS`, call `ParsePkgDescIndexFile` with that FS, and verify the parsed entries.

7. Test the missing-file case using the same in-memory filesystem simply by requesting a path that is not present.

8. Preserve useful error semantics. In particular, `cmdPkgDescIndexVerify` should report the friendly “pkg_desc_index not found ... run generate first” message only when the underlying error is `fs.ErrNotExist` / equivalent. Do not turn permission errors or other I/O failures into “not found”.

9. Keep `ParsePkgDescIndex(io.Reader)` as the low-level parser. Do not duplicate parsing logic.

10. Do not add txtar for this small two-entry parser test. The fixture is simple enough for `fstest.MapFS`; txtar would be unnecessary complexity here.

11. Avoid introducing another exported API solely for tests if optional injection into the existing API solves the problem cleanly. An unexported helper is fine if it materially improves the implementation.

Run formatting, relevant unit tests, and the full existing test/CI-equivalent suite.

Before finishing, search for all `ParsePkgDescIndexFile` call sites and make sure:

* existing one-argument callers remain source-compatible,
* the real production path uses the injectable filesystem implementation,
* the new unit test exercises that same path without touching disk.

Do not revert the filesystem abstraction again on the basis that it is “only needed for testing”; making side effects injectable is the point of this change.

PR #429, opened 2026-08-09
fix(lint): gracefully handle missing md5-cache for PG0802

arran4:
This fixes an issue where the missing md5-cache lint rule `PG0802` was overly strict. If the repository does not have the `metadata/md5-cache` directory, `g2 lint` will now skip enforcing the rule unless explicitly mandated via the QA policy. A unit test is also provided to assert this behavior.

---
*PR created automatically by Jules for task [16440172670214438005](https://jules.google.com/task/16440172670214438005) started by @arran4*

### DIFF
--- a/lints/md5cache/md5cache_missing_test.go
+++ b/lints/md5cache/md5cache_missing_test.go
@@ -9,8 +9,19 @@ import (
 	"github.com/arran4/g2/lints"
 )
 
+type mockFileStat struct {
+	paths map[string]bool
+}
+
+func (m mockFileStat) Stat(name string) (os.FileInfo, error) {
+	if _, ok := m.paths[name]; ok {
+		return nil, nil // Return a fake file info, not used by code anyway.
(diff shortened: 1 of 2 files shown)


lints/md5cache/md5cache_missing_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/arran4/g2"
+	"github.com/arran4/g2/lints"
+)
+
+func TestMD5CacheMissing(t *testing.T) {
+	tempDir := t.TempDir()
```
arran4: @jules refactor to move file operations into an interface so that we can avoid doing file system changes or access in tests

arran4: @jules refactor to move file operations into an interface so that we can avoid doing file system changes or access in tests

PR #448, opened 2026-08-16
⚡ Optimize ebuild deduplicate hash logic

arran4:
💡 **What:** The optimization stream lines directly to the MD5 hasher within the parsing loop using `io.WriteString` instead of accumulating them into a slice to join later.
🎯 **Why:** Reduces memory allocations and avoids the overhead of creating a slice and joining strings before hashing.
📊 **Measured Improvement:** The benchmark results confirm that the new logic significantly reduces allocations per operation, providing an improvement in memory usage and execution time.

---
*PR created automatically by Jules for task [2308209894000412344](https://jules.google.com/task/2308209894000412344) started by @arran4*

### DIFF
--- a/ebuild_deduplicate.go
+++ b/ebuild_deduplicate.go
@@ -4,6 +4,7 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


ebuild_deduplicate.go
```diff
@@ -76,14 +77,19 @@ func DeduplicateEbuilds(targets []string) ([]string, error) {
...
+		hasher := md5.New()
 		slot := "0"
+		firstLine := true
 		for _, line := range lines {
 			trimmed := strings.TrimSpace(line)
 			if strings.HasPrefix(trimmed, "# Generated via:") {
 				continue
 			}
-			hashContent = append(hashContent, line)
+			if !firstLine {
```
arran4: @jules don't just testing against the line number do the same thing>

arran4: @jules not the query.. I think first line could be replaced with using `lineNumber` like variable from the range

PR #457, opened 2026-08-18
feat(lint): add `g2 lint commits` subcommand

arran4:
Added `g2 lint commits` to detect and lint only packages modified locally and in unpushed git commits.

---
*PR created automatically by Jules for task [12120509884282558821](https://jules.google.com/task/12120509884282558821) started by @arran4*


arran4:
Blocker is me.. I gotta make decisiosn and understand

arran4:
@jules 
# Prompt: Reimplement incremental `g2 lint changed`

Implement incremental linting against the current `g2` codebase.

The goal is to let a developer efficiently lint only the Gentoo packages affected by their current Git work, without changing the behaviour of the existing lint engine.

The earlier implementation used the name `g2 lint commits`, but its actual behaviour combined committed branch changes with staged, unstaged, and untracked files. Use the clearer command:

`g2 lint changed [flags] [<location>]`

Keep this as a thin target-selection layer over the existing lint implementation. Do not duplicate lint execution or filtering logic.

Requirements:

1. Determine changed repository-relative paths robustly. Include staged and unstaged tracked changes and untracked non-ignored files. Do not run a redundant cached diff when the chosen Git operation already includes staged changes.

2. Also support committed changes on the current branch relative to a meaningful base. Add `--base <ref>` so callers and CI can explicitly select that base. When `--base` is omitted, use the branch's configured upstream when available. Use merge-base semantics so changes introduced on the current branch are selected rather than unrelated changes made to the base branch.

3. If there is no configured upstream and no explicit `--base`, do not arbitrarily fall back to `HEAD~1`. In that situation, lint the working-tree/index/untracked changes only. Make the behaviour clear in help text.

4. Treat Git failures correctly. A directory that is not a Git work tree, an invalid explicit base ref, or another genuine Git error must produce an actionable error rather than silently reporting that there are no modified packages.

5. Parse Git filenames safely. Prefer NUL-delimited Git output so whitespace, unusual characters, renames, etc. cannot corrupt path parsing. Handle deleted and renamed files deliberately.

6. Convert changed paths to lint targets only when they actually identify Gentoo package paths. Do not interpret paths such as `metadata/layout.conf`, `profiles/repo_name`, or other repository infrastructure as category/package pairs. Determine whether repository-level changes should cause applicable repository lints to run; otherwise ignore them deliberately and test/document that behaviour.

7. Feed the resulting package/repository selection into the existing `runLintCore`/lint selection machinery so output formats, severity/source/tag filters, disabled rules, pkgcheck integration, and exit semantics remain consistent with the other `g2 lint` subcommands.

8. Add focused tests covering at least: clean repository; unstaged modification; staged modification; untracked package file; committed branch change with upstream; explicit `--base`; no upstream; invalid base; deleted package file; renamed package file; filename containing whitespace; repository-level metadata/profile change; and invocation outside a Git work tree.

9. Keep output deterministic where observable and keep Git-discovery logic in a small independently testable helper rather than embedding substantial subprocess logic in the CLI handler.

10. Update command help/documentation as appropriate. Remove any accidental/generated files unrelated to the feature.

Run the full existing Go tests, vet/lint checks, and the repository's relevant CI-equivalent checks. Investigate any existing `Go lint/test` failure and fix it if the feature causes or exposes it; do not paper over unrelated failures.

Keep the implementation narrowly scoped. Preserve existing `g2 lint`, `g2 lint repo`, `g2 lint package`, and `g2 lint query` behaviour.

At the end, summarize the exact `g2 lint changed` semantics, how the default base is selected, what happens without an upstream, and the tests added.


Also still pondering about this

arran4:
@jules
Please finish the `g2 lint changed` implementation from the previous iteration.

The overall design is now correct, so keep `g2 lint changed`, `--base`, upstream/merge-base semantics, NUL-delimited Git output, and reuse of `runLintCore`.

There are a few correctness issues remaining:

1. Fix changed-path classification.

Do not infer a Gentoo package merely because a path has 2 or 3 slash-separated components.

The current implementation can incorrectly interpret paths such as:

- .github/workflows/ci.yml
- eclass/foo.eclass
- licenses/GPL-2
- scripts/foo/bar

as package names.

Use the repository's actual Gentoo package/category information to decide whether a changed path belongs to a lintable package.

Prefer validating candidate `category/package` values against the repository model/categories rather than maintaining a growing exclusion list.

A change under a real package such as:

    app-misc/foo/foo-1.ebuild
    app-misc/foo/metadata.xml
    app-misc/foo/Manifest

must select `app-misc/foo`.

Repository infrastructure changes that do not correspond to a package should not manufacture package targets.

Document/test the intended behaviour for metadata/, profiles/, eclass/, licenses/, .github/, etc.

2. Do not ignore Git command failures.

The current calls equivalent to:

    _ = runZCmd("git", "diff", ...)
    _ = runZCmd("git", "ls-files", ...)

must propagate actionable errors.

A failed Git command must not be interpreted as "there are no changed packages".

Keep the special handling where absence of a configured upstream is normal and simply means there is no committed-change base unless --base was supplied.

3. Make returned package lists deterministic.

Sort the resulting package names before returning them.

4. Improve the tests.

The existing "rename/add" test is only an addition. Add a genuine rename test using git mv (ideally including movement between package paths where meaningful).

Also add focused cases for:

- completely clean repository
- staged modification
- unstaged modification
- untracked package file
- actual rename
- deletion
- explicit --base
- upstream base
- no upstream
- invalid base
- whitespace/special filename
- metadata/layout.conf
- profiles/*
- eclass/*
- licenses/*
- .github/*
- deterministic result ordering
- Git command/error behaviour where practical

Avoid one giant stateful test if splitting these into table-driven/subtests or small independent repositories makes failures easier to understand.

5. Remove the unrelated generated/empty:

    cmd/g2/metadata/pkg_desc_index

It must not be part of this change.

6. Ensure help text makes the default semantics clear:

- working tree/index/untracked package changes are always considered;
- committed changes are compared to --base when supplied;
- otherwise the configured upstream is used when present;
- without either, only local working-tree/index/untracked changes are considered.

Keep the feature narrowly scoped. Do not duplicate lint execution logic or change the established lint result/filter semantics.

Run gofmt, the relevant unit tests, go test ./..., go vet ./..., and repository lint checks.

At the end, summarize what changed and specifically confirm the non-package path cases no longer become bogus package targets.

arran4:
@jules
Please finish the current `g2 lint changed` implementation.

Most of the previous review is now addressed. Keep the existing Git semantics, --base behaviour, NUL-delimited output, Git error propagation, sorted results, help text, and runLintCore integration.

There is one important architectural issue remaining.

`cmdLintChanged` currently calls:

    parseRepo(...)

to obtain SiteData purely so getGitModifiedPackagesChanged can validate categories.

It then calls:

    runLintCore(...)

which performs another full parseRepo.

Do not parse the entire overlay twice. `parseRepo` performs substantial repository/package/ebuild parsing, so this undermines the purpose of a fast incremental lint command.

Refactor changed-path classification so it is lightweight and happens before the normal lint parse.

The Git discovery helper should:

1. Discover changed paths using the current Git behaviour.
2. Derive possible category/package paths.
3. Validate those candidates cheaply against the repository filesystem/repository layout rather than requiring a full SiteData parse.
4. Return only actual package candidates.
5. Then invoke runLintCore once; let the existing lint machinery perform its normal repository parse exactly once.

For example, a candidate category/package should correspond to an actual Gentoo package directory in the current repository. Do not treat arbitrary paths underneath a legitimate category as packages.

Consider deleted/renamed paths deliberately:
- a modified/deleted file in an existing package should select that package;
- a rename should select the current destination package where applicable;
- a completely deleted package cannot be linted from the current working tree and should not cause a bogus package target or failure.

Keep infrastructure paths such as these out of the package target set:

    metadata/*
    profiles/*
    eclass/*
    licenses/*
    .github/*
    scripts/*

but achieve this through positive package identification rather than an ever-growing hard-coded exclusion list.

Add/adjust tests to demonstrate:
- a legitimate package path is selected;
- a nonexistent package name under a legitimate category is not selected;
- metadata/profiles/eclass/licenses/.github paths are ignored;
- a genuine rename works;
- deleted-file behaviour is deliberate;
- returned targets remain deterministic.

Also remove the unrelated empty/generated file:

    cmd/g2/metadata/pkg_desc_index

It must not remain in this change.

Do not duplicate or redesign the lint engine. The end state should be:

    Git discovery / cheap target selection
        ->
    one invocation of runLintCore
        ->
    one normal repository parse and lint execution

Run gofmt, the focused tests, go test ./..., go vet ./..., and the repository lint checks.

At the end, confirm explicitly that `g2 lint changed` no longer performs a preliminary full parseRepo before runLintCore.

PR #508, opened 2026-08-20
feat: add overlay mask and unmask commands

arran4:
This adds `g2 overlay mask` and `g2 overlay unmask` commands to allow modifying the `profiles/package.mask` and `profiles/package.unmask` configuration files.

These commands support four main subcommands:
- `list`: lists all entries
- `check`: checks if a specific package is masked/unmasked
- `add`: adds new packages
- `remove`: removes existing packages

This commit mirrors the behavior of `g2 package masked` but within the context of an overlay.

---
*PR created automatically by Jules for task [11624096924147365680](https://jules.google.com/task/11624096924147365680) started by @arran4*


arran4:
@jules 
Rework the current branch substantially. The existing implementation should **not** be merged in its current form.

Do not assume access to PR/issue metadata or rebase context; work from the requirements below and the code currently on this branch.

## Primary correction

The current implementation adds:

```text
g2 overlay mask ...
g2 overlay unmask ...
```

and modifies:

```text
profiles/package.mask
profiles/package.unmask
```

That is the wrong abstraction for the repo-scoped local Portage policy feature.

The intended canonical interface is:

```text
g2 conf overlay <repo> ...
```

This command manages the **local machine's Portage configuration for a selected repository**, not files inside the overlay repository.

For example:

```text
g2 conf overlay arrans-overlay list

g2 conf overlay arrans-overlay mask app-misc/foo
g2 conf overlay arrans-overlay unmask app-misc/foo

g2 conf overlay arrans-overlay mask-reset app-misc/foo
g2 conf overlay arrans-overlay unmask-reset app-misc/foo
```

These operations should produce repository-qualified atoms such as:

```text
app-misc/foo::arrans-overlay
```

in the appropriate local Portage configuration.

## Do not modify repository profile files

Remove the new behaviour that mutates:

```text
profiles/package.mask
profiles/package.unmask
```

This feature must instead operate on:

```text
/etc/portage/package.mask
/etc/portage/package.unmask
```

or equivalent paths beneath `--config-root`.

Do not introduce `profiles/package.unmask` as an overlay-maintainer mechanism.

If editing an overlay's own `profiles/package.mask` is desirable, that is a separate feature and should not be implemented as part of this work.

## Repository resolution

`g2 conf overlay <repo>` must resolve `<repo>` using configured `repos.conf`.

Requirements:

* support normal `/etc/portage/repos.conf` configuration, including directory-form configuration where existing repo parsing already supports it;
* distinguish the configured repository name from its filesystem basename;
* verify that the requested repository exists/configured before mutating anything;
* use the actual repository name when constructing `::repo` atoms;
* reject ambiguous repository names rather than silently guessing;
* support:

  * `--config-root`
  * `--repos-conf`
    for testing and non-default installations.

If existing `g2` helpers already perform repository resolution, reuse them rather than introducing another implementation.

## Mask/unmask semantics

Keep mask and unmask as distinct policy layers.

For:

```text
g2 conf overlay arrans-overlay mask app-misc/foo
```

add the semantic equivalent of:

```text
app-misc/foo::arrans-overlay
```

to local `package.mask`.

For:

```text
g2 conf overlay arrans-overlay unmask app-misc/foo
```

add the semantic equivalent to local `package.unmask`.

Do **not** implement unmasking by removing a mask rule automatically.

Likewise, reset operations should explicitly distinguish:

```text
mask-reset
unmask-reset
```

and remove only the corresponding user configuration.

If the caller supplies an atom that already contains a repository qualifier:

```text
app-misc/foo::arrans-overlay
```

accept it if it agrees with the selected repository.

If it conflicts:

```text
g2 conf overlay arrans-overlay mask app-misc/foo::guru
```

return a clear error. Do not silently rewrite the qualifier.

## Reuse existing mask-management infrastructure

Do not duplicate the current `package_mask.go` parser into another `package_unmask.go` parser unless there is genuinely no suitable shared abstraction.

The repository already has mask-management work. Refactor toward reusable primitives for package configuration where practical.

In particular, avoid duplicating:

* parsing;
* serialization;
* matching;
* mutation;
* source/provenance handling;
* flat-file vs directory handling.

`package.mask` and `package.unmask` contain essentially the same kind of atom lines from the user's configuration-management point of view, so shared machinery is preferable.

## Preserve configuration exactly where possible

The current implementation is unsafe because it parses a subset of a file and then rewrites the entire file using `os.Create`.

Do not do that.

A valid existing file such as:

```text
# My local masks
>=cat/foo-2
cat/bar
```

must not lose its entries or comments simply because it does not follow a special:

```text
# Name <email> (date)
```

header format.

Mutations must preserve unrelated:

* comments;
* blank lines;
* existing atoms;
* ordering;
* whitespace where practical;
* independently managed fragments.

Do not require author/date metadata for ordinary `/etc/portage/package.mask` or `/etc/portage/package.unmask` rules.

## Flat-file and directory forms

Handle both:

```text
/etc/portage/package.mask
/etc/portage/package.mask/*
```

and:

```text
/etc/portage/package.unmask
/etc/portage/package.unmask/*
```

without changing the user's chosen layout.

Rules:

* if the target is a flat file, edit that flat file safely;
* if the target is a directory, use a predictable g2-owned fragment such as:

```text
package.mask/g2.conf
package.unmask/g2.conf
```

or an equally clear existing convention;

* do not silently convert a flat file into a directory;
* do not collapse a directory of fragments into one file;
* do not delete arbitrary user-owned fragments.

An explicit `--file` override may be supported if it fits the existing command model.

## Safe writes

Do not truncate the live file before knowing that the replacement is valid.

Use an atomic/safe mutation strategy where practical, for example:

1. construct the complete replacement;
2. write it to a temporary file in the same filesystem;
3. close/sync as appropriate;
4. rename over the target.

Errors must leave the previous configuration intact.

## Atom matching

Remove uses of:

```go
strings.Contains(entry.Package, pkgName)
```

for semantic package checks.

This produces false positives such as:

```text
cat/foo
cat/foobar
```

and cannot understand Gentoo atoms.

Use the existing Gentoo atom parser/matcher if available.

The implementation should support normal atom forms expected by the rest of `g2`, including as applicable:

```text
cat/pkg
=cat/pkg-1.2
>=cat/pkg-2
cat/pkg:0
cat/pkg::repo
```

Do not invent a second incomplete atom matcher.

## Idempotence

Adding an existing semantic rule should not create duplicates.

For example:

```text
g2 conf overlay arrans-overlay mask app-misc/foo
g2 conf overlay arrans-overlay mask app-misc/foo
```

should leave one equivalent rule and report that the desired state already exists.

Reset/removal should similarly be safe when the rule is already absent.

## Listing

Implement:

```text
g2 conf overlay list
```

to list configured repositories.

Implement:

```text
g2 conf overlay <repo> list
```

to show user policy relevant to that repository, at least:

```text
Masks:
Unmasks:
```

with useful source provenance where the existing parser supports it.

The listing should discover repository-qualified entries regardless of which fragment filename contains them.

Do not assume that repo-specific rules live only in a file named after the repository.

## Scope

Do not attempt to implement keyword acceptance in this change unless the necessary package.accept_keywords infrastructure is already present on the branch and extending the command cleanly is trivial.

The important result here is to establish the correct:

```text
g2 conf overlay <repo>
```

architecture and correct reusable mask/unmask behavior.

Leave room for later commands such as:

```text
keywords
keywords-all
keywords-reset
enable
disable
status
```

without hard-coding a command structure that makes those awkward.

## Broad repo operations

If you decide to include repo-wide mask/unmask operations in this branch, use:

```text
*/*::<repo>
```

and make them explicit operations such as:

```text
mask-all
unmask-all
```

Do not infer them from package-level commands.

`unmask-all` is dangerous because it can override repository/profile masks protecting broken, vulnerable or incompatible packages.

If implemented now, it must require an explicit acknowledgement such as:

```text
--force
```

and should support `--dry-run`.

It is acceptable to leave these operations for follow-up work instead of expanding this branch.

## Tests

Add meaningful tests for the command and mutation behavior, not merely parser entry counts.

At minimum cover:

1. repository resolution from a temporary `repos.conf`;
2. package atom automatically receives the selected `::repo` qualifier;
3. matching existing correct repo qualifier is accepted;
4. conflicting repo qualifier is rejected;
5. mask adds to local `package.mask`;
6. unmask adds to local `package.unmask`;
7. mask-reset affects only mask configuration;
8. unmask-reset affects only unmask configuration;
9. repeated add is idempotent;
10. repeated reset is idempotent;
11. unrelated existing entries survive mutation;
12. comments and blank lines survive mutation;
13. a headerless valid Portage file survives mutation;
14. flat-file configuration works;
15. directory/fragment configuration works;
16. unrelated fragments remain untouched;
17. exact atom matching does not confuse `cat/foo` with `cat/foobar`;
18. versioned/operator atoms survive correctly;
19. repository-qualified atoms survive correctly;
20. simulated write failure does not truncate the existing configuration, where practical to test.

Prefer filesystem-backed integration tests using a temporary `--config-root` and `--repos-conf` so the tests exercise the actual command behavior without touching the host system.

## Cleanup

Remove code added by the current implementation that is no longer appropriate, including duplicated overlay mask/unmask command files and duplicated unmask parsing structures where replaced by shared machinery.

Run formatting, unit tests, lint/static analysis, and the repository's normal test suite.

The finished branch should represent a coherent implementation of repo-scoped **local Portage mask/unmask policy**, not a command for rewriting an overlay's own profile files.

arran4:
@jules
Continue improving the code on the current branch after commit `569ecfa`.

The architectural rework is now broadly correct: keep the canonical interface as:

```text
g2 conf overlay <repo> ...
```

and keep mutations in local Portage configuration rather than modifying files inside the overlay repository.

Do not revert to `g2 overlay mask/unmask` or `profiles/package.mask` mutation.

There are lint failures plus several functional issues that need to be fixed before this is ready.

## 1. Fix all current lint failures

Current lint reports:

```text
cmd/g2/conf_overlay.go:244:14: Error return value of os.MkdirAll is not checked
cmd/g2/conf_overlay.go:305:14: Error return value of f.Close is not checked
cmd/g2/conf_overlay.go:306:16: Error return value of os.Remove is not checked
cmd/g2/conf_overlay.go:318:14: Error return value of os.Remove is not checked
cmd/g2/conf_overlay_test.go:16:20: Error return value of os.RemoveAll is not checked
cmd/g2/conf_overlay_test.go:89:14: Error return value of os.MkdirAll is not checked
cmd/g2/conf_overlay_test.go:90:15: Error return value of os.WriteFile is not checked
cmd/g2/conf_overlay_test.go:135:10: Error return value of w.Close is not checked
cmd/g2/conf_overlay_test.go:138:15: Error return value of buf.ReadFrom is not checked
cmd/g2/conf_overlay_test.go:97:12: ineffectual assignment to err
cmd/g2/conf_overlay_test.go:109:12: ineffectual assignment to err
```

Fix these properly rather than suppressing linters.

For tests, prefer `t.TempDir()` rather than manually calling `os.MkdirTemp` + `RemoveAll`.

Check all filesystem/read/write/pipe errors in the new tests, including ones not currently reported because they are explicitly assigned to `_`.

Cleanup failures may use `_ = ...` where cleanup failure genuinely cannot affect the primary result, but normal operations should return/wrap useful errors.

## 2. Fix `g2 conf overlay <repo> list`

The current list implementation is incorrect.

The mutation code writes ordinary Portage configuration such as:

```text
app-misc/foo::test-repo
```

without an author/date header.

However, `cmdConfOverlayList` currently calls `parsePackageMaskDir`, which eventually uses the old `g2.ParsePackageMasked` parser. That parser only records atom lines after its special Gentoo repository-mask author header:

```text
# Name <email> (date)
```

As a result, ordinary `/etc/portage/package.mask` and `package.unmask` entries — including entries this new command creates itself — can disappear from `list`.

Do not use the repository `profiles/package.mask` parser as the parser for ordinary user Portage package configuration.

Implement or reuse a simple provenance-preserving user package-config reader suitable for:

```text
/etc/portage/package.mask
/etc/portage/package.mask/*

/etc/portage/package.unmask
/etc/portage/package.unmask/*
```

It must recognize normal non-comment atom lines regardless of surrounding comments.

It should retain at least:

* atom text;
* source file;
* preferably source line number.

Then make:

```text
g2 conf overlay test-repo list
```

discover all matching `::test-repo` rules across arbitrary fragments.

Do not require rules to live in `g2.conf`.

Do not silently suppress unexpected parsing/read errors. Missing optional configuration is fine; permission/I/O errors should be returned clearly.

## 3. Make the list test actually test the result

The current `list outputs` test has:

```go
if buf.String() != expected {
    fmt.Printf(...)
}
```

This does not fail the test.

Change mismatches to `t.Errorf`/`t.Fatalf`.

Also make stdout capture robust:

* check `os.Pipe()` errors;
* check `w.Close()`;
* check `buf.ReadFrom()` or preferably use an appropriate helper;
* restore `os.Stdout` reliably even when the test fails.

Add a regression test proving that a headerless file such as:

```text
# local policy
app-misc/foo::test-repo
app-misc/bar::another-repo
```

causes the first atom, and only the first atom, to appear in:

```text
g2 conf overlay test-repo list
```

Test both flat-file and directory-fragment forms.

## 4. Reject conflicting repository qualifiers consistently

Mask/unmask additions already reject:

```text
g2 conf overlay test-repo mask app-misc/foo::guru
```

But `mask-reset` and `unmask-reset` currently don't perform the same validation.

That means a command scoped to `test-repo` can currently remove a rule belonging to `guru`.

Centralize repository qualification/validation into one helper and use it for:

* `mask`
* `unmask`
* `mask-reset`
* `unmask-reset`

Desired behaviour:

```text
app-misc/foo
    -> app-misc/foo::test-repo

app-misc/foo::test-repo
    -> accepted unchanged

app-misc/foo::guru
    -> clear error
```

Add tests for conflicting qualifiers on both reset commands.

Avoid implementing this validation four times independently.

## 5. Implement `g2 conf overlay list`

There are two distinct list forms:

```text
g2 conf overlay list
g2 conf overlay <repo> list
```

The first should list configured repositories.

The second should show policy for one repository.

Refactor `cmdConfOverlay` parsing so `list` can be recognized before assuming the first argument is a repository name.

For the all-repositories form, read configured, enabled repositories from the selected `repos.conf`.

Provide `--repos-conf`/`--config-root` testability in a way consistent with the rest of this command family.

Exact output formatting can remain simple, but write tests for the command.

## 6. Tighten repository resolution

Do not silently use `/var/db/repos` as a fallback when repository resolution is supposed to come from a supplied `repos.conf`.

In particular, this must behave deterministically:

```text
g2 conf overlay foo ... --repos-conf /tmp/test/repos.conf
```

If `foo` is not configured there, it should not unexpectedly succeed just because `/var/db/repos/foo` happens to exist on the host running the command or tests.

Repository selection should be based on configured `repos.conf`.

Also keep the distinction between:

* repos.conf section/configuration identity;
* repository filesystem location;
* actual Portage repository name.

Use the actual repository identity for generated:

```text
::<repo>
```

qualification.

If the configured location contains `profiles/repo_name`, use/reuse the appropriate existing mechanism where available rather than assuming the section name and repository name can never differ.

Add tests for:

* disabled repository;
* missing repository;
* directory-form `repos.conf`;
* explicit test `--repos-conf` not falling through to host `/var/db/repos`;
* differing configured identity/repository name if supported by the existing repository model.

## 7. Improve mutation safety

The original destructive whole-file rewrite has been removed, which is good, but make both add and remove mutations robust.

### Adds

`appendToConfig` currently writes directly with `O_APPEND`.

Prefer a safe mutation mechanism that constructs the intended replacement and atomically replaces the file, while preserving unrelated bytes/content.

At minimum, do not leave a partially appended atom if a write fails.

### Removes

Do not use a predictable:

```text
<filename>.tmp
```

temporary path.

Use `os.CreateTemp` in the same directory so rename remains atomic and concurrent invocations do not share a temp filename.

Preserve relevant metadata of an existing config file, especially its file mode.

Ensure every error path closes and removes temporary files appropriately without introducing further errcheck failures.

Keep comments, blank lines and unrelated entries unchanged.

## 8. Preserve idempotence

Retain the current useful behaviour where adding the exact existing rule does not append another copy.

Add explicit tests:

```text
mask foo
mask foo
```

results in one rule.

Likewise repeated:

```text
mask-reset foo
mask-reset foo
```

must succeed safely.

Cover the same behaviour for unmask.

If suitable atom-normalization/matching infrastructure already exists in this repository, reuse it rather than introducing another ad-hoc parser. Otherwise keep the current scope conservative and exact rather than pretending string substring matching is semantic atom matching.

## 9. Improve test isolation

The current subtests share one mutable Portage tree and depend on actions performed by earlier subtests.

Refactor where practical so tests set up the state they require explicitly rather than relying on subtest execution order.

Use small helpers for:

* creating repos.conf;
* reading required files with error checking;
* invoking commands;
* capturing stdout.

Tests should fail for incorrect behaviour rather than merely printing diagnostics.

## 10. Run the full validation suite

After fixing the above:

* run `gofmt` on changed Go files;
* run the repository's normal unit tests;
* run the same lint/staticcheck/errcheck suite used by CI;
* fix all new warnings rather than suppressing them.

Do not stop once the 11 currently reported lint messages disappear; ensure tests then exercise the functional issues above.

The desired result is a clean, testable implementation of repo-scoped local mask/unmask policy under `g2 conf overlay`, with reliable listing, deterministic repository resolution, consistent qualifier validation, and safe filesystem mutation.

arran4:
This PR is being superseded by a clean replacement implementation because the automated iteration stopped making effective changes.

arran4:
Superseded by #513 (clean implementation on `feat/conf-overlay-mask-policy`).

PR #511, opened 2026-08-20
fix: support main gentoo categories even if overlay restricts categories

arran4:
Overlays like Guru often provide a restricted `profiles/categories` list containing only their custom categories. They rely on the main portage tree to define the standard categories (like `dev-util`). Previously, `g2` site generation would skip valid directories if they were not listed in the overlay's custom categories list, leading to missing packages on the static site. This change allows directories that are part of the main Gentoo categories list to be processed.

---
*PR created automatically by Jules for task [5673616112108073467](https://jules.google.com/task/5673616112108073467) started by @arran4*


arran4:
why...

PR #516, opened 2026-08-21
lints/news: Enforce single values for Display-If-* headers

arran4:
Added checks to verify that `Display-If-Installed`, `Display-If-Keyword`, and `Display-If-Profile` headers in news items don't contain spaces or tabs, preventing users from putting multiple values on one line. Updated the corresponding tests in `lints/news/news_test.go` to assert these cases trigger lint failures as expected.

---
*PR created automatically by Jules for task [655187163272658827](https://jules.google.com/task/655187163272658827) started by @arran4*

### DIFF
--- a/lints/news/news.go
+++ b/lints/news/news.go
@@ -346,13 +346,27 @@ func (r *NewsValidityLintRule) lintNewsItem(content string, relPath string) []li
 				results = append(results, res)
 			}
 		case "Display-If-Installed":
-			if !strings.Contains(val, "/") {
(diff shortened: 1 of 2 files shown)


lints/news/news.go
arran4: @jules should we move to a: 
```
type Validator func(detailsandstuff) error
var (
  validators = map[string /*header key*/][]Validator {
     // ...
  }
)
```

Format, b/c of repetition etc?

PR #517, opened 2026-08-21
Refactor news-item header validation

arran4:
Replaces the large switch block in `lintNewsItem` with a map-driven architecture utilizing a local state struct (`NewsItemState`) and `HeaderContext` wrapper to improve code maintainability and testability. Covered with new table-driven tests.

---
*PR created automatically by Jules for task [12011874756618852029](https://jules.google.com/task/12011874756618852029) started by @arran4*

### DIFF
--- a/lints/news/news.go
+++ b/lints/news/news.go
@@ -203,14 +203,7 @@ func (r *NewsValidityLintRule) LintWithQA(repoDir string, pkg *g2.PackageData, q
 }
 
 type NewsItemState struct {
-	HasTitle       bool
-	HasAuthor      bool
-	HasPosted      bool
-	HasRevision    bool
-	HasFormat      bool
-	HasContentType bool
-	NewsItemFormat string
-	ContentTypeVal string
+	Headers map[string]string
 }
 
 type HeaderContext struct {
@@ -333,50 +326,24 @@ func validateCategoryPackage(ctx *HeaderContext, state *NewsItemState) []lints.L
 
 var headerValidators = map[string][]HeaderValidator{
 	"Title": {
-		func(ctx *HeaderContext, state *NewsItemState) []lints.LintResult {
-			state.HasTitle = true
-			return nil
-		},
 		validateTitleLength,
 	},
 	"Author": {
-		func(ctx *HeaderContext, state *NewsItemState) []lints.LintResult {
-			state.HasAuthor = true
-			return nil
-		},
 		validateEmailFormat,
 	},
 	"Translator": {
 		validateEmailFormat,
 	},
 	"Posted": {
-		func(ctx *HeaderContext, state *NewsItemState) []lints.LintResult {
-			state.HasPosted = true
-			return nil
-		},
 		validatePostedDate,
 	},
 	"Revision": {
-		func(ctx *HeaderContext, state *NewsItemState) []lints.LintResult {
-			state.HasRevision = true
-			return nil
-		},
 		validateRevisionInteger,
 	},
 	"Content-Type": {
-		func(ctx *HeaderContext, state *NewsItemState) []lints.LintResult {
-			state.HasContentType = true
-			state.ContentTypeVal = ctx.Value
-			return nil
-		},
 		validateContentType,
 	},
 	"News-Item-Format": {
-		func(ctx *HeaderContext, state *NewsItemState) []lints.LintResult {
-			state.HasFormat = true
-			state.NewsItemFormat = ctx.Value
-			return nil
-		},
 		validateNewsItemFormat,
 	},
 	"Display-If-Installed": {
@@ -396,7 +363,7 @@ func (r *NewsValidityLintRule) lintNewsItem(content string, relPath string) []li
 	lines := strings.Split(content, "\n")
 
 	inBody := false
-	state := &NewsItemState{}
+	state := &NewsItemState{Headers: make(map[string]string)}
 
 	for lineNum, line := range lines {
 		if inBody {
@@ -437,6 +404,8 @@ func (r *NewsValidityLintRule) lintNewsItem(content string, relPath string) []li
 		key := strings.TrimSpace(parts[0])
 		val := strings.TrimSpace(parts[1])
 
+		state.Headers[key] = val
+
 		ctx := &HeaderContext{
 			Key:     key,
 			Value:   val,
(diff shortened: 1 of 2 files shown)


lints/news/news.go
```diff
@@ -202,6 +202,26 @@ func (r *NewsValidityLintRule) LintWithQA(repoDir string, pkg *g2.PackageData, q
 	return results
 }
 
+type NewsItemState struct {
```
arran4: @jules perhaps we should be passing a map rather than "Has" 

`map[string][]string` I guess or it can be part of the field as "Headers"

lints/news/news.go
```diff
@@ -282,126 +404,57 @@ func (r *NewsValidityLintRule) lintNewsItem(content string, relPath string) []li
...
-		default:
+		} else {
 			res := lints.LintResult{RuleMetadata: ruleNewsValidity, Message: fmt.Sprintf("[%s] Unknown header: '%s'", lints.SeverityWarning, key), File: relPath, Line: lineNum + 1}
 			res.RuleMetadata.Severity = lints.SeverityWarning
 			results = append(results, res)
 		}
 	}
 
-	if !hasTitle {
+	if _, ok := state.Headers["Title"]; !ok {
```
arran4: @jules all of these ifs should be done by the validators, we can modify the validators to return seveirty and other things as required to make this work

arran4:
@jules Please continue PR #517 and finish the news-item header validation refactor with a smaller, more declarative validation API.

The current implementation is better than the original switch, but it still has too much ceremony: `HeaderContext`, `NewsItemState`, separate `PostValidator` plumbing, repeated `LintResult` construction, and tests that require complete news items to exercise individual header rules.

Refactor toward this model:

```go
type Headers map[string][]string

type Finding struct {
    Severity lints.Severity
    Message  string
}

type HeaderValidator func(key string, values []string) []Finding
type Validator func(headers Headers) []Finding
```

Exact names may differ if there is a clearer local naming convention, but preserve the design principles below.

The header registry should be concise and declarative, approximately:

```go
var headerValidators = map[string][]HeaderValidator{
    "Title": {
        required(),
        single(),
        nonEmpty(),
        maxLength(50),
    },
    "Author": {
        required(),
        each(validEmail()),
    },
    "Translator": {
        each(validEmail()),
    },
    "Posted": {
        required(),
        single(),
        each(validDate("2006-01-02")),
    },
    "Revision": {
        required(),
        single(),
        each(validInteger()),
    },
    "Content-Type": {
        single(),
        each(validContentType()),
    },
    "News-Item-Format": {
        required(),
        single(),
        each(validNewsItemFormat()),
    },
    "Display-If-Installed": {
        each(validInstalledPackage()),
    },
    "Display-If-Keyword": {
        each(validKeyword()),
    },
    "Display-If-Profile": {
        each(validProfile()),
    },
}
```

Do not add a `Required bool` field. Requiredness is itself validation. The validation engine should iterate over the declared header registry, including headers that are absent from the parsed input, so `required()` receives zero values and can report the missing-header finding.

Use `map[string][]string`, not `map[string]string`, so repeated headers are preserved. GLEP 42 permits multiple Author and Translator headers, and repeated Display-If-* fields are meaningful.

Keep ordinary header validators independent of file paths, line numbers, `LintResult`, `NewsValidityLintRule`, or a general context/state object. They should accept only the header key and collected values and return semantic `Finding` values.

Use a separate very small `Validator func(headers Headers) []Finding` only for genuinely cross-header rules, such as the News-Item-Format 1.0 / Content-Type requirement. Do not make every per-header validator receive the complete header collection just for uniformity.

Centralise conversion of `Finding` to `lints.LintResult` in one helper. That helper should:

* copy `ruleNewsValidity`
* set the finding severity
* add the existing `[Severity]` prefix
* attach the appropriate file and line information

Validators should not repeatedly construct `LintResult` or mutate `RuleMetadata.Severity`.

Preserve existing externally visible lint messages and behaviour unless fixing a demonstrable regression.

Be careful with dependent validation. In particular, do not implement `Display-If-Installed` as two independent validators that both fire on the same malformed value and introduce cascading diagnostics. Preserve the old first-failure behaviour. A value like:

```text
Display-If-Installed: foo bar
```

should produce the multiple-values/spaces diagnostic, not also an additional category/package diagnostic caused by continuing to validate structurally broken input.

For that reason, prefer a semantic validator such as:

```go
each(validInstalledPackage())
```

where the internal checks happen in the appropriate order, rather than composing independent validators when later checks depend on earlier structural validity.

Apply the same principle elsewhere: use small reusable validators/combinators where they improve readability, but do not build a generic validation DSL or abstraction framework for the entire repository.

The parser/orchestration path should become simple:

1. Parse headers and preserve every value in `Headers`.
2. Retain enough source-location information separately to attach line numbers to findings arising from parsed headers.
3. Run the declared `HeaderValidator`s for every registered header, including absent ones.
4. Run the cross-header validators.
5. Convert findings to `LintResult` centrally.
6. Continue existing body validation.

Improve the tests at the same time.

Individual header behaviour should be directly testable without constructing complete news-item documents. Add concise helpers so tests can look conceptually like:

```go
testHeader(t, "Title", nil,
    errorFinding("Missing required header: Title"))

testHeader(t, "Title", []string{""},
    errorFinding("Title cannot be empty"))

testHeader(t, "Author", []string{"Invalid Author"},
    errorFinding("Invalid Author format, expected 'Name <email@domain.com>': 'Invalid Author'"))

testHeader(t, "News-Item-Format", []string{"3.0"},
    warningFinding("Unsupported News-Item-Format: '3.0'"))
```

Compare findings structurally and exactly, including severity and message. Unexpected extra findings must fail the test. Do not use substring matching as the primary assertion mechanism, and do not merely log result-count mismatches.

Add explicit tests for:

* missing required headers via `required()`
* valid and invalid Title values
* Author and Translator email validation
* multiple valid Author/Translator occurrences
* date, revision and format validation
* repeated Display-If-* values
* `Display-If-Installed: foo bar` producing only the appropriate first structural diagnostic
* News-Item-Format 1.0 requiring Content-Type text/plain
* unknown headers
* preservation of source file/line information when findings are converted to `LintResult`

Keep only a small number of full `lintNewsItem` end-to-end tests covering parser integration, required headers, cross-header behaviour, repeated headers, unknown headers and body validation. Remove repeated full valid-news boilerplate where it only exists to reach one individual header validator.

Do not generalise this into repository-wide lint infrastructure in this PR.

Run gofmt, targeted tests, the complete Go test suite, and the repository's normal lint/static-analysis checks. Fix failures rather than weakening assertions.

arran4:
@jules Please make one final focused pass on PR #517 after commit `1480718`.

The declarative validator API is now substantially where it should be. Do not redesign it again broadly. Fix the remaining integration and regression issues.

### 1. Preserve the source line of individual repeated header occurrences

The current implementation stores:

```go
lineNumbers map[string]int
```

which records only the first occurrence of each header.

This is incorrect now that `Headers` deliberately preserves multiple values. For example:

```text
Author: Valid Person <valid@example.com>
Author: invalid
```

an error for the second Author must point at the second Author line, not the first.

Keep file/absolute line information out of the semantic value validators themselves. Use a small mechanism that lets a finding identify the header occurrence that produced it.

For example, an occurrence/value index on a finding is reasonable:

```go
type Finding struct {
    Severity   lints.Severity
    Message    string
    // optional occurrence/index information for per-value findings
}
```

and `each()` is a natural place to associate returned findings with the value index.

The orchestration layer can then translate that occurrence index through separately collected source locations when creating the final `LintResult`.

The exact representation is up to you; keep it small. Do not reintroduce `HeaderContext` or make ordinary semantic validators accept file paths and line numbers.

Add a test with multiple Authors where only the second Author is invalid and assert the exact resulting `LintResult.Line`.

### 2. Make lint output deterministic

The current implementation ranges over both:

```go
headerValidators
headers
```

as maps. Go map iteration order is unspecified, so diagnostics such as missing required headers and unknown headers can appear in varying orders.

Lint output should be stable.

Prefer making the validator registry itself ordered rather than creating a second manually-maintained list of the same header names.

For example, something conceptually like:

```go
type HeaderRule struct {
    Key        string
    Validators []HeaderValidator
}

var headerRules = []HeaderRule{
    header("Title", required(), single(), nonEmpty(), maxLength(50)),
    header("Author", required(), each(validEmail())),
    // ...
}
```

would retain a concise declarative API while defining deterministic validation order.

Names/helpers can differ.

If a lookup structure is useful for detecting unknown headers, derive it from the ordered rules rather than maintaining two independent declarations that can drift.

Unknown-header diagnostics should also have deterministic ordering, preferably source order.

Avoid otherwise changing lint messages or semantics.

### 3. Restore coverage accidentally removed during the test cleanup

Moving individual header rules to direct validator tests was correct.

However, the previous `TestNewsValidityLintRule` also covered behaviour outside header validation which has now disappeared from the test suite in this PR, including:

* invalid news directory names
* invalid news filenames
* filename/directory prefix mismatch
* body tab-character validation
* body 72-character wrapping validation

Restore focused coverage for those behaviours.

Do not restore the giant old all-in-one test fixture. Prefer small targeted tests for the directory/file validation and body validation paths.

Keep the new direct header tests.

### 4. Keep the current good properties

Do not undo these parts of `1480718`:

* `required()` remains a validator, not a `Required bool`
* repeated header values remain preserved
* Author/Translator and Display-If-* can have repeated occurrences where valid
* `Display-If-Installed` keeps first-failure semantics so `foo bar` does not cascade into a second category/package diagnostic
* semantic validators return `Finding`s rather than constructing `LintResult`
* `LintResult` construction remains centralised
* cross-header validation remains separate from ordinary per-header validation
* direct validator tests compare findings exactly rather than substring matching

Also review `validDate(format)` while here: it currently accepts a generic format argument but hardcodes `YYYY-MM-DD` in its diagnostic. Either make the diagnostic correspond to the supplied format or make the helper intentionally specific to the Posted date rather than presenting a misleading generic API.

Run gofmt, the targeted news tests, the complete Go test suite, and the repository's normal lint/static-analysis checks.

Keep this pass strictly focused on finishing PR #517.

arran4:
@jules 
Please make one final small fix to PR #517 after `644ce39`.

The overall refactor is now good and CI is green. Do not change the architecture again.

There is still a small source-line attribution hole.

`Finding.ValueIndex` is populated automatically by `each()`, but `nonEmpty()` and `maxLength()` currently iterate over `values` themselves and return findings without recording which value failed.

This means a malformed duplicated Title such as:

```text
Title: Valid title
Title:
```

can report `Title cannot be empty` against the first Title line rather than the second.

Make value-specific validators consistently go through the per-value path.

Prefer making `nonEmpty()` and `maxLength()` value validators and using them through `each()`, conceptually:

```go
header(
    "Title",
    required(),
    single(),
    each(nonEmpty()),
    each(maxLength(50)),
)
```

Adjust exact types/names as appropriate, but do not introduce another context object.

Also improve `single()`: when multiple occurrences exist, its “must appear at most once” finding should point to the first duplicate occurrence (index 1), rather than the original occurrence at index 0.

Add focused tests covering:

```text
Title: Valid title
Title:
```

and a duplicated Title whose second value exceeds the maximum length. Assert that the value-specific finding points to the second Title line.

If useful, also assert that the `single()` finding itself points to the second occurrence.

Keep all existing behaviour, messages, deterministic ordering, repeated-header handling, cross-header validation and restored regression coverage unchanged.

Run gofmt, targeted news tests and the normal repository test/lint suite.

PR #537, opened 2026-08-30
🧪 Test ParseProjectsFromReader read errors

arran4:
Adds focused coverage for the missing `io.Reader` failure path in `ParseProjectsFromReader`.

`ParseProjectsBytes` already delegates directly to `ParseProjectsFromReader`, and its existing tests cover successful XML parsing, malformed XML, invalid root elements, and decoded `Projects` contents. This PR therefore focuses on the behavior that was not previously covered: propagating an error returned while reading from the supplied `io.Reader`.

### Intended coverage

- A reader failure returns `nil` projects and a non-nil error.
- The original reader error remains identifiable with `errors.Is`.
- Existing parsing tests remain the source of coverage for valid XML, malformed XML, and invalid roots rather than duplicating those fixtures.

### Verification

- `go fmt ./...`
- `go test ./...`

---
*PR created automatically by Jules for task [7586977170582937386](https://jules.google.com/task/7586977170582937386) started by @arran4*

### DIFF
--- a/cmd/g2/lint_changed_test.go
+++ b/cmd/g2/lint_changed_test.go
@@ -5,7 +5,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"testing"
-
 )
 
 func TestGetGitModifiedPackagesChanged(t *testing.T) {
@@ -36,9 +35,13 @@ func TestGetGitModifiedPackagesChanged(t *testing.T) {
 
 	// 2. Initial state
 	err = os.MkdirAll(filepath.Join(tmpDir, "app-misc", "foo"), 0755)
-	if err != nil { t.Fatal(err) }
+	if err != nil {
+		t.Fatal(err)
+	}
 	err = os.WriteFile(filepath.Join(tmpDir, "app-misc", "foo", "foo-1.ebuild"), []byte(""), 0644)
-	if err != nil { t.Fatal(err) }
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	runCmd("git", "add", ".")
 	runCmd("git", "commit", "-m", "Initial commit")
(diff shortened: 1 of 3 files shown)


arran4:
@jules Please revise this PR based on review:

- Remove the duplicated valid XML, malformed XML, invalid-root, and full `Projects` fixture cases from the new `TestParseProjectsFromReader`. `TestParseProjectsBytes` already covers those semantics and delegates directly to `ParseProjectsFromReader`, so duplicating them here adds maintenance cost without covering a new branch.
- Keep the new coverage focused on the previously untested `io.Reader` failure path.
- Replace the ad-hoc `fmt.Errorf("read error")` reader with an injected sentinel error, e.g. a reader struct carrying an `err error` field and returning that from `Read`.
- Assert that `ParseProjectsFromReader` returns `nil` for the result and that the reader error is preserved/identifiable with `errors.Is(err, sentinelErr)`. Do not merely assert that some error occurred.
- Remove the XML-name-space normalization and `reflect.DeepEqual` machinery from this new test; they should no longer be necessary once the duplicated parsing cases are removed.
- Run `go fmt ./...` so the import/order and extra whitespace are normalized.
- Run `go test ./...` after the change.

Please keep this test-only; no production change is needed. I have updated the PR title and description separately to reflect the narrower intended scope.

arran4:
@jules The substantive test fix now looks good. Please make one cleanup pass before this is merged:

- Revert the unrelated `go fmt ./...` churn in `cmd/g2/lint_changed_test.go`.
- Revert the unrelated whitespace-only changes in `use_desc_test.go`.
- Keep the focused `projects_test.go` changes exactly as intended: sentinel reader error, `errors.Is` verification, and `nil` result assertion.
- After reverting the unrelated files, run `gofmt` only on the file(s) actually changed for this PR as needed, then run `go test ./...`.

The final PR diff should ideally contain only the focused `projects_test.go` change.

PR #538, opened 2026-09-03
ci: fix multiple GitHub release creators by enforcing GoReleaser as sole owner

arran4:
### Previous release-owner graph

`workflow_dispatch release-*` / tag push → `goreleaser` → `publish-draft` → `promote-release`

This allowed more than one component to own GitHub Release creation. GoReleaser could create the release while `softprops/action-gh-release` attempted to create a competing draft for the same tag, leaving duplicate/orphan release state.

### New release-owner graph

GoReleaser is the sole GitHub Release publisher. `publish-draft`, `promote-release`, and the incomplete `prepare-next-version-pr` job are removed.

- **Manual stable / RC / alpha release:** `prepare-release-tag` calculates and validates the intended semantic tag only → required Go lint/test gate must succeed → `goreleaser` creates/pushes that exact tag idempotently and verifies an existing tag points to `github.sha` → GoReleaser publishes the release in the same workflow run. The tag push uses `GITHUB_TOKEN`, so it does not create a duplicate workflow run.
- **Manual `release-test`:** calculates the test version → required Go test gate must succeed → GoReleaser runs with `--snapshot`. No permanent tag is pushed and no GitHub Release is published.
- **Externally pushed `v*` tag:** `prepare-release-tag` is intentionally skipped → required Go test gate must succeed → GoReleaser publishes the tagged commit.
- **Externally pushed `test-*` tag:** required Go test gate must succeed → GoReleaser runs in snapshot mode and does not publish a GitHub Release.

### Safety properties

- A failed or skipped `go-test` cannot enter the release job.
- Manual permanent tags are created only after the test gate succeeds.
- Manual tag creation is retry-safe: an existing tag is reused only when it resolves to the exact current `github.sha`; a conflicting tag fails loudly.
- RC and alpha modes are normal prerelease publication paths, not snapshots.
- Test modes remain non-publishing snapshots.
- There is one GitHub Release owner: GoReleaser.

### Files changed

- `.github/workflows/ci.yml`
- `.github/workflows/validate_release_graph.py`

### Validation

- Workflow/YAML validation and the Go test suite passed during development.
- PR CI exercises a dedicated `Validate Release Graph Policy` job which statically verifies the manual `workflow_dispatch release-*` path, external `v*` and `test-*` tag paths, snapshot versus publishing semantics, failed/skipped test gating, single publisher ownership, and retry-safe SHA-checked manual tagging.
- At head `4f1af02`, `Validate Release Graph Policy` passed in CI run 2110. The latest commit only tightens validation assertions for the manual release entry paths.

### Follow-up outside this PR

Historical untagged or duplicate draft releases created by the previous competing publisher paths should be reviewed and cleaned up manually.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,3 +1,8 @@
+# Agent rules for generation:
+# https://arran4.github.io/blog/post/2026/042-simplified-github-ci-release-safe/
+# Release safety details:
+# https://arran4.github.io/blog/post/2026/041-release-safe-single-owner-github-ci/
+
 name: CI/CD
 
 on:
@@ -228,6 +233,10 @@ jobs:
           clean_tag="${next_tag#v}"; clean_tag="${clean_tag%%-*}"
           IFS='.' read -r maj min pat <<< "$clean_tag"
           echo "next_version=${maj:-0}.${min:-0}.$(( ${pat:-0} + 1 ))-SNAPSHOT" >> "$GITHUB_OUTPUT"
+
+          # Create and push the tag to trigger the actual release pipeline
+          git tag "$next_tag"
+          git push origin "$next_tag"
   gitleaks:
     name: Secret scan
     needs: [route, discover]
@@ -485,8 +494,8 @@ jobs:
   goreleaser:
     name: GoReleaser
     # In practice, include all quality gates here (for example: go-test, go-vet, go-lint, format).
-    needs: [route, discover, go-test, prepare-release-tag]
-    if: ${{ needs.discover.outputs.has_go == 'true' && needs.discover.outputs.has_goreleaser == 'true' && (((github.event_name == 'push') && startsWith(github.ref, 'refs/tags/v')) || (github.event_name == 'workflow_dispatch' && startsWith(inputs.mode, 'release-'))) }}
+    needs: [route, discover, go-test]
+    if: ${{ needs.discover.outputs.has_go == 'true' && needs.discover.outputs.has_goreleaser == 'true' && (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')) }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -495,9 +504,6 @@ jobs:
       - uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
-      - name: Tag commit locally
-        if: ${{ github.event_name == 'workflow_dispatch' && startsWith(inputs.mode, 'release-') }}
-        run: git tag ${{ needs.prepare-release-tag.outputs.release_tag }}
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v6
         with:
@@ -505,42 +511,13 @@ jobs:
           version: '~> v2'
           args: >-
             release --clean
-            ${{ (github.event_name == 'workflow_dispatch' && (inputs.mode == 'release-test' || inputs.mode == 'release-rc' || inputs.mode == 'release-alpha')) && '--snapshot' || '' }}
+            ${{ contains(github.ref, '-test') && '--snapshot' || '' }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAP_GITHUB_TOKEN: ${{ secrets.TAP_GITHUB_TOKEN }} # inject secrets.TAP_GITHUB_TOKEN
-          GORELEASER_CURRENT_TAG: ${{ needs.prepare-release-tag.outputs.release_tag }}
-  publish-draft:
-    name: Publish draft release assets
-    needs:
-      - goreleaser
-    if: ${{ needs.route.outputs.run_release == 'true' }}
-    runs-on: ubuntu-latest
-    steps:
-      - name: Collect artifacts
-        uses: actions/download-artifact@v4
-        with:
-          path: dist-release
-      - name: Publish draft GitHub release
-        uses: softprops/action-gh-release@v2
-        with:
-          draft: true
-          files: dist-release/**
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-  promote-release:
-    name: Promote draft to published
-    needs: [publish-draft]
-    if: ${{ github.event_name == 'release' || (github.event_name == 'workflow_dispatch' && startsWith(inputs.mode, 'release-')) }}
-    runs-on: ubuntu-latest
-    steps:
-      - name: Release promoted via upstream process
-        run: echo "Promotion step placeholder (gh api patch release draft=false)"
-
   prepare-next-version-pr:
     name: Prepare next development iteration PR
-    needs: [publish-draft]
+    needs: [prepare-release-tag]
     if: ${{ github.event_name == 'workflow_dispatch' && (startsWith(inputs.mode, 'release-') || inputs.mode == 'release-test') }}
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -536,3 +536,18 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAP_GITHUB_TOKEN: ${{ secrets.TAP_GITHUB_TOKEN }} # inject secrets.TAP_GITHUB_TOKEN
           GORELEASER_CURRENT_TAG: ${{ needs.prepare-release-tag.outputs.release_tag || '' }}
+
+  release-graph-validation:
+    name: Validate Release Graph Policy
+    needs: [route]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+      - name: Install PyYAML
+        run: pip install pyyaml
+      - name: Run release graph validation
+        run: python3 .github/workflows/validate_release_graph.py
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -496,7 +496,7 @@ jobs:
       always() &&
       needs.discover.outputs.has_go == 'true' &&
       needs.discover.outputs.has_goreleaser == 'true' &&
-      (needs.go-test.result == 'success' || needs.go-test.result == 'skipped') &&
+      (needs.go-test.result == 'success') &&
       (needs.prepare-release-tag.result == 'success' || needs.prepare-release-tag.result == 'skipped') &&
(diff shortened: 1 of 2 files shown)


arran4:
@jules Please fix the release workflow architecture before this PR is considered ready.

There is a critical trigger bug in the current implementation: `prepare-release-tag` uses `actions/checkout`'s persisted repository `GITHUB_TOKEN` credentials and then runs `git push origin "$next_tag"`. GitHub intentionally does **not** create a new workflow run for ordinary `push` events caused by `GITHUB_TOKEN`. Therefore a manual `release-*` dispatch will push the tag, but the new tag-push-only `goreleaser` job will never get a second workflow run and no release will be published.

Please rework this without adding a new PAT/GitHub-App secret unless one is already intentionally part of this repository's release design. Keep GoReleaser as the sole GitHub Release publisher and keep `publish-draft` / `promote-release` removed.

A suitable shape is:

1. Manual `workflow_dispatch` release modes should calculate/create/push the tag and then run the GoReleaser publication path **in the same workflow run**. The `GITHUB_TOKEN` suppression of the resulting tag-push workflow is then expected and prevents a duplicate run.
2. Independently/user-pushed semantic tags should continue to run the same GoReleaser publication job from the normal `push` event.
3. Make the GoReleaser job's `needs`/`if` logic support both paths safely. Be careful that `prepare-release-tag` is skipped on ordinary tag pushes: if it is listed in `needs`, use an explicit `always()`/result-aware condition (or an equivalent structure) so a skipped manual-only dependency does not suppress the tag-push release path.
4. For the manual path, preserve the generated tag as `GORELEASER_CURRENT_TAG` / local tag context as needed, and ensure the release is for the exact checked-out commit.
5. `prepare-next-version-pr` must run only after successful GoReleaser publication for a manual release, not merely after `prepare-release-tag`; otherwise a failed release can still advance the development version.
6. Preserve the intended semantics of `release-test`, `release-rc`, and `release-alpha`; do not accidentally turn a test/snapshot mode into a normal published release. Verify this against the previous workflow behavior and the repository's release policy.
7. Add/adjust validation that checks the event/dependency graph for both cases: (a) manual `release-*` dispatch publishes exactly once and then permits the next-version PR, and (b) an externally pushed release tag publishes exactly once. Existing PR CI is insufficient because both changed release jobs are skipped on `pull_request`.

Please keep the change focused on the release-owner problem, run the applicable workflow/YAML validation and Go tests, and report exactly how the two event paths behave after the fix.

arran4:
@jules `647f990` fixes the earlier `GITHUB_TOKEN` trigger problem, but this is still not release-safe. Please address the following before another review:

1. **Do not push the permanent release tag before the quality gate has passed.** `prepare-release-tag` currently only `needs: [route]`, so it can run `git tag` / `git push origin "$next_tag"` while `golangci` / `go-test` are still running. If tests fail, or the workflow is cancelled after the push, the repository is left with a release tag that has no release; because the push used `GITHUB_TOKEN`, no tag-push workflow will rescue it, and retrying the manual release fails because the tag already exists. Restructure this so manual tag creation/push happens only after the required gates succeed (ideally in or immediately before the sole GoReleaser publication path), and ensure the tagged commit is exactly the dispatch SHA.

2. **Do not broaden tag publishing to every tag accepted by the workflow.** The current GoReleaser condition changed from `startsWith(github.ref, 'refs/tags/v')` to `startsWith(github.ref, 'refs/tags/')`. The workflow itself also accepts `test-*`, so a manually/user-pushed `test-*` tag now enters GoReleaser as a normal release; the snapshot expression does not match `refs/tags/test-*` because it checks for `-test`. Restore a precise release-tag filter (prefer the repository's actual `v...` release-tag grammar), and explicitly define what `test-*` should do rather than letting it publish accidentally.

3. **Correct the snapshot/publication semantics.** GoReleaser `--snapshot` builds locally and does **not** upload/publish artifacts or create a GitHub Release. Therefore a successful `release-test`, `release-rc`, or `release-alpha` snapshot cannot be treated as a successful publication. Re-check the repository's intended policy for those modes. If RC/alpha are intended to be GitHub prereleases, use the appropriate real-release/prerelease mechanism rather than `--snapshot`; if they are intentionally validation-only, do not push a permanent release tag or run the post-release version-advance path as though publication occurred.

4. **Fix or disable the now-activated next-version PR path.** `prepare-next-version-pr` can now receive `next_version`, but the only repo-specific bump command is still a commented placeholder. It will `git add -A` and then `git commit` with no changes, so a successful manual release can end with a deterministic job failure. Either implement the actual g2 version-advance change, or make this job an intentional no-op/remove it for a tag-derived-version repository.

5. **Add the event/dependency validation requested in the previous review.** The current PR CI is green, but `Prepare release tag`, `GoReleaser`, and `Prepare next development iteration PR` were all skipped in that PR run, so it does not validate the release graph. Add a focused static/testable validation covering at least: manual stable release after gates; failed/cancelled gate cannot strand a tag; external valid `v...` tag publishes once; `test-*` cannot publish a normal release; and prerelease/snapshot modes follow the documented policy.

Keep GoReleaser as the sole GitHub Release publisher and keep `publish-draft` / `promote-release` removed. Keep the change focused, run `actionlint`, applicable workflow validation, and `go test ./...`, then report the exact behavior for stable manual, externally pushed stable tag, test, RC, and alpha paths.

arran4:
@jules `d089e38` is closer, but this is **not ready to approve yet**. Please address the remaining release-safety issues:

1. **Do not allow a skipped `go-test` to satisfy the release gate.** The current condition permits `(needs.go-test.result == 'success' || needs.go-test.result == 'skipped')`. `go-test` depends on `golangci`; if lint fails, GitHub can skip `go-test`, and because the GoReleaser job uses `always()`, the release job can still proceed. For release paths, `go-test` is expected to run, so require `needs.go-test.result == 'success'`. Keep the skipped allowance only for `prepare-release-tag`, because that job is intentionally skipped on externally pushed tags.

2. **Make the manual tag/publish path retry-safe.** The tag is now pushed after the test gate, which fixes the earlier pre-gate orphan-tag problem, but it is still pushed immediately before GoReleaser publication. If GoReleaser then fails (transient GitHub/API/TAP/build/publish failure), `Re-run failed jobs` will execute the tag step again and `git tag <same-tag>` / `git push` can fail because the tag already exists. Make the tag step idempotent: if the intended tag already exists locally/remotely, verify that it resolves to the exact `github.sha` and reuse it; fail loudly if it points anywhere else. Only create/push it when absent. Do not silently bump to a new version on retry.

3. **The requested release-graph validation is still missing.** PR CI is green, but on run 2106 both `Prepare release tag` and `GoReleaser` were skipped, so it still does not exercise the changed release path. Add a focused static/logic validation (or equivalent test) covering at least: manual stable release, manual rc/alpha release, manual `release-test` snapshot, externally pushed `v*` tag, externally pushed `test-*` snapshot, and the failed/skipped gate cases. The validation should prove a failed lint/test cannot publish and that each valid path has exactly one publisher.

4. **Update the PR body after the code is fixed.** It is now stale: it still says `prepare-release-tag` pushes the tag and a tag-push event fires GoReleaser, and it still describes `prepare-next-version-pr`. The actual intended graph after `d089e38` is different. Please rewrite the old/new release-owner graph and validation section to match the final implementation, including the non-publishing `release-test` semantics.

The changes in `d089e38` that should be retained are: removing the broken placeholder next-version PR job, limiting snapshot semantics to test modes, publishing rc/alpha normally, and keeping GoReleaser as the sole GitHub Release publisher. Keep the change focused and report the exact behavior of each release mode after the fix.

arran4:
@jules `a3ce72c` fixes the two remaining code defects from the last review: the release gate now requires `needs.go-test.result == 'success'`, and manual tag creation is retry-safe/idempotent with SHA verification. Please retain those changes.

One item is still missing before I consider this ready: **add focused release-graph validation**. The latest PR CI run is green, but `Prepare release tag` and `GoReleaser` are still skipped on the pull-request event, so CI does not exercise the changed release control flow.

Please add a small static/logic validation (script/test/workflow-policy test; no real release publication required) that covers at least:

- manual stable release: test success -> one publisher; failed/skipped test -> no tag/publish
- manual RC and alpha: publish path, not snapshot
- manual `release-test`: snapshot only, no permanent tag/publish
- externally pushed `v*`: `prepare-release-tag` skipped but GoReleaser allowed after successful tests
- externally pushed `test-*`: snapshot only
- conflicting existing manual tag SHA: fail
- existing correct manual tag SHA: retry succeeds without re-creating/re-pushing it

The validation should assert that each valid publishing path has exactly one GitHub Release owner and that failed gates cannot publish. Keep this focused on `.github/workflows/ci.yml` release semantics.

I have updated the PR body myself to match the current implementation, so no PR-description work is needed from you. Run the focused validation plus the normal applicable checks and report the results.

arran4:
@jules `202adad` adds the right validation script, but this is still **not ready to approve** because the validation is not actually executed by CI.

The current PR run 2108 contains no job or step invoking `.github/workflows/validate_release_graph.py`; `Prepare release tag` and `GoReleaser` are still skipped, and the new static check itself never ran. Please wire the script into normal PR CI as a required job/step so regressions fail the PR.

Also tighten the script so it explicitly asserts the complete trigger matrix we asked for, not just portions of the current strings:

1. manual `workflow_dispatch` `release-*` enters `goreleaser` only with `go-test == success`;
2. externally pushed `refs/tags/v*` enters `goreleaser` with `prepare-release-tag` skipped;
3. externally pushed `refs/tags/test-*` enters `goreleaser` in snapshot mode;
4. manual `release-test` is snapshot-only and does not run the permanent tag-push step;
5. manual `release-rc` / `release-alpha` are non-snapshot publishing paths;
6. failed/skipped test gates cannot publish;
7. only GoReleaser owns GitHub Release publication;
8. manual permanent tag creation remains retry-safe and SHA-checked.

A small dedicated `release-graph-validation` job is fine. Since the script imports `yaml`, ensure the runner installs/has PyYAML rather than relying on an undeclared environment dependency. Keep this focused; no release behavior changes are otherwise requested.

Once this check is actually part of PR CI and passes, I expect this PR to be at the approve point.

arran4:
@jules `eae792e` fixes the previous CI-wiring blocker: `Validate Release Graph Policy` now runs in PR CI and passed on run 2109. The release implementation itself still looks correct.

There is one remaining gap in the validator before I consider this fully covered: **Check 1 claims to validate the manual `workflow_dispatch release-*` entry path, but it never actually asserts that `goreleaser.if` contains that trigger.** As written, removing the manual release trigger entirely would still leave `validate_release_graph.py` green, provided the test-gate and tag-push assertions remain.

Please add explicit assertions against `g_if` for both parts of the manual entry condition, e.g. that it contains `github.event_name == 'workflow_dispatch'` and `startsWith(inputs.mode, 'release-')`. This should cover stable, RC, alpha, and test manual modes at the job-entry level; the existing args/tag-step assertions can continue to distinguish snapshot vs permanent-tag behavior.

Also restore/retain an assertion that `prepare-release-tag` itself is restricted to manual `workflow_dispatch release-*` runs, since that was present in the first version of the validator and was removed in `eae792e`.

No release-architecture redesign is requested. Keep the change limited to these validator assertions, run PR CI, and report the `Validate Release Graph Policy` result. Once that passes, this should be ready to approve.

